PR #293, opened 2021-09-16
feat: ui cleaning and start accessibility

pentcle:
changed headings to title case
began adding semantic html to text (eg p tags v vanilla text in a div which is not accessible)
removed bold from some headers (whereas should be weight 700 - but ideally we'll use the font face)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@
 .env.development.local
 .env.test.local
 .env.production.local
-
+.idea
 npm-debug.log*
 yarn-debug.log*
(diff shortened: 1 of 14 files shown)


xgambitox:
@pentcle appreciate your help with this!

For some context, we also changed the font at last moment this week to a temporary free alternative, which also might affected the look and feel, but we are looking for one that suits best than this one for the next weeks.

Will ask the team we consider this details you point out for future changes :)

xgambitox (approved):

PR #319, opened 2021-10-08
feat: Iframe for LedgerLive integration

Majorfi:
As part of Yearn's integration with Ledger, an embeded version of beta.yearn.finance, that will be used in the Ledger Live desktop application, should be done.

> The DApp must run differently if it is called from within an iframe or not. If it is, directly initialize the web3 instance with the iframe-provider. This way, there is no web3modal connect flow at all, and users are instantly connected to the wallet of the parent window.

One way to do this would be to add a full wrapper around the app to handle differently the Onboard wallets and the Ledger Iframe. Instead of that I built a "New" wallet for Onboard that will handle [Ledger's plugin].

Ledger's doc: https://developers.ledger.com/docs/dapp/introduction/#dapp-adaptation-on-your-server
Preview URL: https://platform.apps.ledger.com/app/dapp-browser?mock=true&url=https://beta.major.tax/?embed=true
Ledger's plugin: https://www.npmjs.com/package/@ledgerhq/iframe-provider

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2144,7 +2144,7 @@
     "@ledgerhq/errors" "^5.50.0"
     events "^3.3.0"
 
-"@ledgerhq/iframe-provider@^0.4.2":
+"@ledgerhq/iframe-provider@0.4.2":
   version "0.4.2"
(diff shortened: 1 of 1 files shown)


xgambitox (approved):

turtlemoji (approved):
At code level I think is ok, I just don't like the idea of having an injected iframe for security reasons. Approved

PR #322, opened 2021-10-09
fix: withdraw issue on fantom [WEB-941]

xgambitox:
- Fix issue when trying to withdraw from a fantom vault


turtlemoji (approved):
Approved only if last resort

xgambitox:
Dont have an alterntive for this. Should be fine though since Ethers estimates gas for us in every transaction that does not include it.

PR #364, opened 2021-11-12
feat: hide vault if no deposits [WEB-1020]

xgambitox:
- Add ability to hide vaults if user has no deposits on vaults with corresponding metadata


turtlemoji (commented):
Approved

Issue #373, opened 2021-11-17
NFT Owner Access [WEB-1025]


xgambitox:
**Elevator Pitch**
Yearn NFT owners might have special access to features

**Customer Value**
Community can enjoy personalized custom features

**Definition of Done**
- NFT balances are fetched for current selected address and stored in redux state
- Woofy NFT balance is available in state
- Blue Pill NFT balance is available in state
- State clears on wallet disconnect 

**Design Approach**
- Add `getNftBalance(address)` to `UserService.ts` that fetches both Woofy and Blue Pill NFT balance.
- Add new `user` module in redux store, with an `nft` object to handle key-value pairs for each NFT and balance.
- Create `selectUserWoofyBalance` and `selectUserBluePillBalance` selectors
- Add an effect hook on Layout container to execute actions on wallet change event.

**Additional Context**
[Woofy and The Blue Pill NFTs](https://galaxy.eco/yearn/)

TacosTonight:
@xgambitox Hi can I be assigned this issue?

PR #394, opened 2021-12-05
NFT Owner Access [WEB-1025] #373

TacosTonight:
Fixes [#373 ](https://github.com/yearn/yearn-finance-v3/pull/373)

Definition of Done: 
- [x] NFT balances are fetched for current selected address and stored in redux state
- [x] Woofy NFT balance is available in state
- [x] Blue Pill NFT balance is available in state
- [x] State clears on wallet disconnect


src/core/services/UserService.ts
```diff
@@ -1,17 +1,101 @@
...
+        knownMax++;
+      } catch (error) {
+        moreIdsAvailable = false;
+      }
+    }
+    return knownMax - 1;
+  }
+
+  private generateNftIds(starCount: number): number[] {
+    return Array.from({ length: starCount }, (_, i) => i + 1);
```
xgambitox, line 88: we can use `[...Array(starCount+1).keys()]` since we want to include the base 0

src/core/services/UserService.ts
```diff
@@ -1,17 +1,101 @@
...
+  }
+
+  private generateRepeatedAddresses(starCount: number, address: string): string[] {
+    return Array(starCount).fill(address);
```
xgambitox, line 92: Would also require `Array(starCount+1).fill(address)` now that we are including the zero base index

src/core/services/UserService.ts
```diff
@@ -1,17 +1,101 @@
-import { GetAddressEnsNameProps, UserService, Web3Provider } from '@types';
+import { GetAddressEnsNameProps, UserService, Web3Provider, NftBalances, Config } from '@types';
+import { getContract } from '@frameworks/ethers';
+
+import bluePillNFTAbi from './contracts/bluePillNFT.json';
+import woofyNFTAbi from './contracts/woofyNFT.json';
+import BigNumber from 'bignumber.js';
+import { Contract } from 'ethers';
```
xgambitox, line 7: We have not documented a convention, but try moving external dependencies to top.

src/core/store/modules/user/user.reducer.ts
```diff
@@ -0,0 +1,47 @@
+import { UserState } from '@types';
+import { createReducer } from '@reduxjs/toolkit';
```
xgambitox, line 2: Move external dependency to top

src/core/store/modules/user/user.reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+import { UserActions } from './user.actions';
+
+import { initialStatus } from '@types';
```
xgambitox, line 5: combine both imports from `@types`

src/core/store/modules/user/user.reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+
+    /* -------------------------------------------------------------------------- */
+    /*                                 Fetch Data                                 */
+    /* -------------------------------------------------------------------------- */
+    .addCase(getNftBalance.pending, (state) => {
+      state.statusMap.getNftBalance = { loading: true };
+    })
+    .addCase(getNftBalance.rejected, (state, { error }) => {
+      state.statusMap.getNftBalance = { error: error.message };
```
xgambitox, line 39: On error, we need to reset loading state to `false` too

xgambitox (changes requested):
Really good one 👍 Just some small changes and we are good :)

src/core/services/UserService.ts
TacosTonight: was receiving an enable downleveliteration error that would require an update of the tsconfig.json file, so going with this line instead `Array.from({ length: starCount + 1}, (_, i) => i);`

xgambitox (approved):

PR #403, opened 2021-12-17
Vault historical earnings in underlying token

jclancy93:
image

image

image

image

### DIFF
--- a/src/client/components/app/VaultDetail/VaultDetailPanels.tsx
+++ b/src/client/components/app/VaultDetail/VaultDetailPanels.tsx
@@ -242,6 +242,15 @@ export const VaultDetailPanels = ({
     setSelectedTab(selectedTab);
   };
 
+  const chartDataVisible = chartData?.underlying && chartValue?.underlying && chartValue?.underlying !== '0';
+  const chartValueText = selectedUnderlyingData ? (
+    <>
+      {formatAmount(chartValue?.underlying ?? '0', 2)} {selectedVault?.token?.symbol}
+    </>
+  ) : (
+    <>{formatUsd(chartValue?.usd)}</>
+  );
+
   const handleAddToken = () => {
     const { address, symbol, decimals, icon } = selectedVault.token;
     context?.wallet.addToken(address, symbol, decimals, icon || '');
@@ -335,7 +344,7 @@ export const VaultDetailPanels = ({
         </VaultActions>
       </Row>
 
-      {chartData?.underlying && chartValue?.underlying && chartValue?.underlying !== '0' && (
+      {chartDataVisible && (
         <VaultChart>
           <StyledCardHeaderContainer>
             <StyledCardHeader header={t('vaultdetails:performance-panel.header')} />
@@ -349,15 +358,7 @@ export const VaultDetailPanels = ({
 
           <ChartValueContainer>
             <ChartValueLabel>{t('vaultdetails:performance-panel.earnings-over-time')}</ChartValueLabel>
-            <ChartValue>
-              {selectedUnderlyingData ? (
-                <>
-                  {formatAmount(chartValue?.underlying, 2)} {selectedVault?.token?.symbol}
-                </>
-              ) : (
-                <>{formatUsd(chartValue?.usd)}</>
-              )}
-            </ChartValue>
+            <ChartValue>{chartValueText}</ChartValue>
           </ChartValueContainer>
 
           <StyledLineChart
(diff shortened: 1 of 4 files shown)


src/client/components/app/VaultDetail/VaultDetailPanels.tsx
```diff
@@ -317,18 +317,21 @@ export const VaultDetailPanels = ({
...
-      {chartData && (
+      {chartData && chartValue && (
         <VaultChart>
           <StyledCardHeader header={t('vaultdetails:performance-panel.header')} />
 
           <ChartValueContainer>
             <ChartValueLabel>{t('vaultdetails:performance-panel.earnings-over-time')}</ChartValueLabel>
-            <ChartValue>{formatUsd(chartValue)}</ChartValue>
+            <ChartValue>
+              {formatAmount(chartValue, selectedVault?.token?.decimals)} {selectedVault?.token?.symbol}
```
jclancy93, line 327: I didn't see a great way in `utils/` to format a BN to a certain number of decimal points, can do it manually inline with BN but wanted to see if there was preferred way to do it

dudesahn:
Is there a way that we could allow users to toggle between dollar and underlying token values?

FoxTheSin:
@jclancy93 Hey jack, sorry to bother you on here, are you looped into Yearn Contributors telegram?

jclancy93:
> @jclancy93 Hey jack, sorry to bother you on here, are you looped into Yearn Contributors telegram?

I'm in the Yearn web channel @FoxTheSin

gotzenx (approved):

gotzenx (approved):

src/client/components/app/VaultDetail/VaultDetailPanels.tsx
```diff
@@ -361,7 +361,7 @@ export const VaultDetailPanels = ({
           </ChartValueContainer>
 
           <StyledLineChart
-            chartData={selectedUnderlyingData ? chartData?.underlying : chartData?.usd}
+            chartData={selectedUnderlyingData ? chartData?.underlying ?? [] : chartData?.usd ?? []}
```
gotzenx, line 364: Instead of guarding this to never be undefined, shouldnt we guard it or fix it on the file/line it was exploding? Because from the `VaultDetailPanelsProps` interface we say that `chartData` could be undefined still.
So this line fixes this specific bug in this specific file but will not prevent for the bug to reappear if we use the LineChart component somewhere else and pass chartData as undefined from there, right?

src/client/components/common/Charts/LineChart/index.tsx
```diff
@@ -117,7 +118,15 @@ export const LineChart: FC<LineChartProps> = ({ chartData, tooltipLabel, classNa
           return (
             <StyledTooltip align={isFirstHalf ? 'left' : 'right'}>
               <Text>{tooltipLabel || point.serieId}</Text>
-              <Text>{formatUsd(point.data.yFormatted.toString())}</Text>
+              <Text>
```
turtlemoji, line 121: I think we should create a const in this cases to keep the html clean.

`const customSymbolText = customSymbol ? formatAmount.... : formatUsd...;`

and then just do `<Text>{customSymbolText}</Text>

src/client/components/app/VaultDetail/VaultDetailPanels.tsx
```diff
@@ -317,18 +335,35 @@ export const VaultDetailPanels = ({
...
+            <StyledChartTab active={selectedUnderlyingData === false} onClick={() => setSelectedUnderlyingData(false)}>
+              USD
+            </StyledChartTab>
+          </StyledCardHeaderContainer>
 
           <ChartValueContainer>
             <ChartValueLabel>{t('vaultdetails:performance-panel.earnings-over-time')}</ChartValueLabel>
-            <ChartValue>{formatUsd(chartValue)}</ChartValue>
+            <ChartValue>
+              {selectedUnderlyingData ? (
```
turtlemoji, line 353: I think we should create a const in this cases to keep the html clean like in the example below I commented

src/client/components/app/VaultDetail/VaultDetailPanels.tsx
```diff
@@ -317,18 +335,35 @@ export const VaultDetailPanels = ({
         </VaultActions>
       </Row>
 
-      {chartData && (
+      {chartData?.underlying && chartValue?.underlying && chartValue?.underlying !== '0' && (
```
turtlemoji, line 338: I think we should create a const in this cases to keep the html clean, like
`chartDataVisible = chartData?.underlying && chartValue?.underlying && chartValue?.underlying !== '0'` 

and then just do `{charchartDataVisible tData && (....`

turtlemoji (approved):

gotzenx (approved):

x48-crypto (approved):

PR #418, opened 2021-12-31
Feat: supported assets validation [WEB-1060]

gotzenx:


src/core/services/TransactionService.ts
```diff
@@ -1,14 +1,36 @@
...
+import { Interface } from '@ethersproject/abi';
+import {
+  TransactionService,
+  ExecuteTransactionProps,
+  TransactionResponse,
+  GasService,
+  GasFees,
+  YearnSdk,
+} from '@types';
+import { getContract } from '../frameworks/ethers';
```
xgambitox, line 11: Lint: 
- Separate external imports and internal imports with a newLine
- import getContract using paths like `@frameworks/ethers`

src/core/services/IronBankService.ts
```diff
@@ -94,31 +94,48 @@ export class IronBankServiceImpl implements IronBankService {
     const { MAX_UINT256 } = this.config;
     const provider = this.web3Provider.getSigner();
     const ironBankMarketContract = getContract(marketAddress, ironBankMarketAbi, provider);
-
     switch (action) {
       case 'supply':
-        return await this.transactionService.execute({ network, fn: ironBankMarketContract.mint, args: [amount] });
+        return await this.transactionService.execute({
```
FiboApe, line 99: what do you think about having different methods for each one of this switch cases?

executeTransaction ends up being very generic, if you split it up in methods:
- You can get advantage of TS typing by making sure you are always sending the correct parameters
- You make methods here more atomic and thus more testable

src/core/services/IronBankService.ts
```diff
@@ -94,31 +94,48 @@ export class IronBankServiceImpl implements IronBankService {
...
+          network,
```
FiboApe, line 100: Network seems to be a user setting that is not changed based on every call right?

The IronBankService could have an internal variable network that gets updated through a setNetwork and that only gets called on the action that the user changes the network through the UI, this reduces the boilerplate of having to send it all the time.

Does marketAddress work in the same way or would marketAddress also change depending on which action the user is taking?

src/core/services/IronBankService.ts
```diff
@@ -94,31 +94,48 @@ export class IronBankServiceImpl implements IronBankService {
...
+          methodName: 'mint',
+          abi: ironBankMarketAbi,
+          contractAddress: marketAddress,
+          args: [amount],
+        });
       case 'withdraw':
         if (amount === MAX_UINT256) {
           const balanceOf = await ironBankMarketContract.balanceOf(userAddress);
```
FiboApe, line 108: Isn't userAddress derivable from the web3Provider directly instead of having to ask it for every call?

src/core/services/TransactionService.ts
FiboApe: This could probably be made into a eslint rule right?

xgambitox: Ideally yes, but not sure if possible since probably eslint cant distinguish between them. We should at least create a guideline in readme if not possible to enforce via eslint/prettier.

src/core/services/IronBankService.ts
gotzenx: Do agree on that approach (and we pretty much use it all over the repo) but since this piece of code is not gonna be on the FE for too long (since we are gonna move this logic to the SDK) we went for the simplest/generic approach in here.

xgambitox:
I would rename the title as it no longer validates contracts, and is more of a transaction service refactor.

xgambitox (approved):

turtlemoji (approved):
Lgtm

PR #420, opened 2022-01-02
Notification Banner [WEB-1032]

### LINKED ISSUE #372: Notification Banner [WEB-1032]
**Elevator Pitch**
App should display important news/events through a page banner

**Customer Value**
Users are informed with custom messages directly through the site

**Definition of Done**
- Banner component is rendered in all app routes.
- Banner supports 'info' | 'warning' | 'critical' states.
- Banner displays the top most non-dismissed active message available.
- Message can be dismissed by user to display next message.
- Messages support markdown.

**Design Approach**
Create `src/client/data/notificationMessages.json`, with the following structure:
```
[
  id: number,
  type: 'info' | 'warning' | 'critical',
  active: boolean,
  message: 'Welcome to <b>Yearn Finance<b>'
]
```
Add new notifications  module to store, with
- `getNotificationMessages()`, imports json file and filters active messages. Update state in reducer based on users persisted dismissed messages.
- `dismissMessage(messageId)`, sets message status to dismissed.

Include notifications module to `persistConfig` in `core/frameworks/redux`  to persist state on localstorage.
Create a Banner functional component in `common/components` that supports receiving a children component.
Create a NotificationBanner component in `containers`, and use a selector for accessing the first non-dismissed active message. Use `Markdown`  common component.
Import container to all main routes.

**Additional Context**
UI/UX will be provided once a contributor starts a draft PR.

TacosTonight:
Fixes #372 
Added Notifications to redux store

### DIFF
--- a/src/client/data/notificationMessages.json
+++ b/src/client/data/notificationMessages.json
@@ -1,0 +1,23 @@
+[
+    {
+        "id": 0, 
+        "type": "info",
+        "active": true,
+        "message": "Welcome to <b> Yearn Finance 0 <b>" 
+    },
+
+    {
+        "id": 1, 
+        "type": "info", 
+        "active": false, 
+        "message": "Welcome to <b> Yearn Finance 1 <b>"
+    },
+
+    {
+        "id": 2, 
+        "type": "info", 
+        "active": true, 
+        "message": "Welcome to <b> Yearn Finance 2 <b>"
+    }
+
+]
--- a/src/core/frameworks/redux/index.ts
+++ b/src/core/frameworks/redux/index.ts
@@ -3,7 +3,12 @@ import { createLogger } from 'redux-logger';
 import { save, load, clear } from 'redux-localstorage-simple';
 import { merge, cloneDeep, get } from 'lodash';
 
-import rootReducer, { themeInitialState, walletInitialState, settingsInitialState } from '@store/modules';
+import rootReducer, {
+  themeInitialState,
+  walletInitialState,
+  settingsInitialState,
+  notificationsInitialState,
+} from '@store/modules';
 import { enableDevTools } from '@utils';
 import { DIContainer } from '@types';
 
@@ -12,10 +17,11 @@ export const getStore = (extraArgument?: any) => {
     theme: cloneDeep(themeInitialState),
     wallet: cloneDeep(walletInitialState),
     settings: cloneDeep(settingsInitialState),
+    notifications: cloneDeep(notificationsInitialState),
   };
   const persistConfig = {
     namespace: 'yearn',
-    states: ['theme', 'wallet.name', 'settings', 'network'],
+    states: ['theme', 'wallet.name', 'settings', 'network', 'notifications.dismissedMessages'],
   };
   const logger = createLogger({ collapsed: true });
   const middlewareOptions = {
--- a/src/core/store/modules/notifications/notifications.actions.ts
+++ b/src/core/store/modules/notifications/notifications.actions.ts
@@ -1,0 +1,95 @@
+import { createAction, createAsyncThunk } from '@reduxjs/toolkit';
+import { ThunkAPI } from '@frameworks/redux';
+import { keyBy, isEqual, remove } from 'lodash';
+
+import { Message, NotificationsStatus } from '@types';
+import notificationMessages from '../../../../client/data/notificationMessages.json';
+
+/* -------------------------------------------------------------------------- */
+/*                                 Clear State                                */
+/* -------------------------------------------------------------------------- */
+
+const clearActiveMessages = createAction<void>('notifications/clearActiveMessages');
+const clearDismissedMessages = createAction<void>('notifications/clearDismissedMessages');
+
+/* -------------------------------------------------------------------------- */
+/*                                 Fetch Data                                 */
+/* -------------------------------------------------------------------------- */
+const importedNotifications = notificationMessages as Message[];
+
+const getNotificationMessages = createAsyncThunk<NotificationsStatus, void, ThunkAPI>(
+  'notifications/getNotificationMessages',
+  async () => {
+    const activeMessages = importedNotifications.filter((message) => {
+      return message.active === true;
+    });
+    //test arrays are available at end of file
+    const dismissedMessages = removeUpdatedMessagesFromDismissed(testActive, testDismissed);
+    return { active: testActive, dismissed: testDismissed };
+  }
+);
+
+const removeUpdatedMessagesFromDismissed = (activeMessages: Message[], dismissedMessages: Message[]) => {
+  /* 
+  This function handles the scenario where existing messages in notificationMessages.json were updated, 
+  and a user had already dismissed the message. The updated message is removed from dismissedMessages, so
+  that the user can see it. For example if a message was changed from 'warning' to 'critical' then that message 
+  would be removed from dismissedMessages and shown again to the user. 
+  */
+
+  if (!dismissedMessages.length) {
+    return dismissedMessages;
+  }
+
+  const dismissedById = keyBy(dismissedMessages, 'id');
+  for (const [key] of Object.entries(dismissedById)) {
+    const currentActiveMessage = activeMessages.find((message) => message.id === parseInt(key));
+    const currentDismissedMessage = dismissedMessages.find((message) => message.id === parseInt(key));
+    if (isEqual(currentActiveMessage, currentDismissedMessage) === false) {
+      remove(dismissedMessages, (x) => x.id === parseInt(key));
+    }
+  }
+  return dismissedMessages;
+};
+
+/* -------------------------------------------------------------------------- */
+/*                                   Exports                                  */
+/* -------------------------------------------------------------------------- */
+
+export const NotificationsActions = {
+  getNotificationMessages,
+  clearActiveMessages,
+  clearDismissedMessages,
+};
+
+//For testing
+const testDismissed: Message[] = [
+  {
+    id: 0,
+    type: 'warning',
+    active: true,
+    message: 'Welcome to <b> Yearn Finance 0 <b>',
+  },
+  {
+    id: 2,
+    type: 'info',
+    active: true,
+    message: 'Welcome to <b> Yearn Finance 2 <b>',
+  },
+];
+
+//Since type of message id 0 is different, then removed from dismissed state
+const testActive: Message[] = [
+  {
+    id: 0,
+    type: 'critical',
+    active: true,
+    message: 'Welcome to <b> Yearn Finance 0 <b>',
+  },
+  {
+    id: 2,
+    type: 'info',
+    active: true,
+    message: 'Welcome to <b> Yearn Finance 2 <b>',
+  },
+];
--- a/src/core/store/modules/notifications/notifications.reducer.ts
+++ b/src/core/store/modules/notifications/notifications.reducer.ts
@@ -1,0 +1,48 @@
+import { createReducer } from '@reduxjs/toolkit';
+
+import { NotificationsActions } from './notifications.actions';
+import { initialStatus, NotificationsState } from '@types';
+
+export const notificationsInitialState: NotificationsState = {
+  activeMessages: [],
+  dismissedMessages: [],
+  statusMap: {
+    getNotificationMessages: { ...initialStatus },
+  },
+};
+
+const { clearActiveMessages, clearDismissedMessages, getNotificationMessages } = NotificationsActions;
+
+const notificationsReducer = createReducer(notificationsInitialState, (builder) => {
+  builder
+
+    /* -------------------------------------------------------------------------- */
+    /*                                 Clear State                                */
+    /* -------------------------------------------------------------------------- */
+    .addCase(clearActiveMessages, (state) => {
+      state.activeMessages = [];
+    })
+
+    .addCase(clearDismissedMessages, (state) => {
+      state.dismissedMessages = [];
+    })
+
+    /* -------------------------------------------------------------------------- */
+    /*                                 Fetch Data                                 */
+    /* -------------------------------------------------------------------------- */
+    .addCase(getNotificationMessages.pending, (state) => {
+      state.statusMap.getNotificationMessages = { loading: true };
+    })
+
+    .addCase(getNotificationMessages.rejected, (state, { error }) => {
+      state.statusMap.getNotificationMessages = { loading: false, error: error.message };
+    })
+
+    .addCase(getNotificationMessages.fulfilled, (state, { payload: notifications }) => {
+      state.activeMessages = notifications.active;
+      state.dismissedMessages = notifications.dismissed;
+      state.statusMap.getNotificationMessages = {};
+    });
+});
+
+export default notificationsReducer;
--- a/src/core/store/modules/notifications/notifications.selectors.ts
+++ b/src/core/store/modules/notifications/notifications.selectors.ts
@@ -1,0 +1,11 @@
+import { RootState } from '@types';
+
+/* ---------------------------------- State --------------------------------- */
+const selectDismissedMessages = (state: RootState) => state.notifications.dismissedMessages;
+const selectActiveMessages = (state: RootState) => state.notifications.activeMessages;
+
+/* --------------------------------- Exports -------------------------------- */
+export const NotificationsSelectors = {
+  selectDismissedMessages,
+  selectActiveMessages,
+};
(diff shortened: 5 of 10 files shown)


src/client/containers/NotificationBanner/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export const NotificationBanner: FC = () => {
+  const latestMessage = first(useAppSelector(NotificationsSelectors.selectActiveMessages));
+  const dispatch = useAppDispatch();
+
+  let messageSymbol = '';
+  if (latestMessage) {
+    switch (latestMessage.type) {
+      case 'info':
+        messageSymbol = 'ⓘ';
```
FiboApe, line 61: https://github.com/yearn/yearn-finance-v3/pull/424

Is adding an info icon svg to use a React component, maybe we should wait to use that?

src/client/containers/NotificationBanner/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        break;
+      case 'warning':
+        messageSymbol = '⚠';
```
FiboApe, line 64: Should we use svg icons here instead of unicode?

src/client/containers/NotificationBanner/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        messageSymbol = '⚠⚠⚠';
+        break;
+      default:
+        messageSymbol = '';
+    }
+  }
+
+  return (
+    <StyledNotificationBanner>
+      {latestMessage && (
```
FiboApe, line 76: if latestMessage does not exist you could just return null for this component and avoid having to do this check here

src/client/components/common/Banner/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const StyledBanner = styled.div`
+  background: ${({ theme }) => theme.colors.surface};
+  position: fixed;
+  height: 16px;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  z-index: 9999;
```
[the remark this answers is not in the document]
FiboApe: as a side note, not sure a notification banner should be floating at the top of the screen. Most of the full-width banners our there actually occupy space on the DOM.

Wdyt?

src/client/data/notificationMessages.json
```diff
@@ -0,0 +1,30 @@
+[
```
FiboApe, line 1: Probably want to remove this ones for the final merge

src/core/store/modules/notifications/notifications.actions.ts
```diff
@@ -0,0 +1,37 @@
...
+import notificationMessages from '../../../../client/data/notificationMessages.json';
+
+const dismissMessage = createAction<number>('notifications/dismissMessage');
+
+/* -------------------------------------------------------------------------- */
+/*                                 Clear State                                */
+/* -------------------------------------------------------------------------- */
+
+const clearMessages = createAction<void>('notifications/clearMessages');
+const clearDismissedMessageIds = createAction<void>('notifications/clearDismissedMessageIds');
```
FiboApe, line 14: Is this actually ever going to be used? Doesn't seem like it would be a user-facing option and I cannot see a case where we would actually like to clear the user dismissed messages

src/core/store/modules/notifications/notifications.actions.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/* -------------------------------------------------------------------------- */
+/*                                 Fetch Data                                 */
+/* -------------------------------------------------------------------------- */
+const importedNotifications = notificationMessages as Message[];
+
+const getNotificationMessages = createAsyncThunk<Message[], void, ThunkAPI>(
+  'notifications/getNotificationMessages',
+  async () => {
+    return importedNotifications;
```
FiboApe, line 24: Probably here we should also clear on dismissedMessagesId's the ids that are no longer active/existing, else the state could grow endlessly until a user clears the app local storage and it could produce perf issues for that user

src/client/components/common/Banner/index.tsx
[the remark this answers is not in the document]
xgambitox: Will ask on TG for some help with design itself, since issue did not define clear requirements for styling.

src/client/containers/NotificationBanner/index.tsx
```diff
@@ -0,0 +1,91 @@
+import { FC } from 'react';
+import { first } from 'lodash';
+import styled from 'styled-components';
+import { TransitionGroup, CSSTransition } from 'react-transition-group';
+import { useAppSelector, useAppDispatch } from '@hooks';
+import { NotificationsActions, NotificationsSelectors } from '@store';
+import { Banner, Markdown, Icon, CloseIcon } from '@src/client/components/common';
```
xgambitox, line 7: We have to still write a guideline for this:
- imports should follow this order: external imports -> internal imports using path alias-> internal imports using relative paths (all separated by new line)
- prefer directly importing from the direct path if its available. For `@src/client/components/common` do `@components/common`. See tsconfig.paths.json

src/client/containers/NotificationBanner/index.tsx
xgambitox: yes, lets try using svg icons for this ones.

src/client/data/notificationMessages.json
xgambitox: yes, dont forget about this one

src/core/store/modules/notifications/notifications.actions.ts
```diff
@@ -0,0 +1,37 @@
+import { createAction, createAsyncThunk } from '@reduxjs/toolkit';
+import { ThunkAPI } from '@frameworks/redux';
+
+import { Message } from '@types';
+import notificationMessages from '../../../../client/data/notificationMessages.json';
```
xgambitox, line 5: use path alias import '@client/data...'

src/core/store/modules/notifications/notifications.actions.ts
xgambitox: Lets filter them on the reducer, and just add the active messages to state. Still dont compare with user dismissed messages though, since you do that already on selector. Intention should be this list will be short anyways, and we can always trim it.

src/core/store/modules/notifications/notifications.actions.ts
xgambitox: Yeah, probably not needed as of now, so lets delete them so we can maintain only used code.

xgambitox (changes requested):
Made some comments, good overall! 

If you can add some styling on the banner to git it more space would be better. Will ask if someone can help with designs.

src/core/store/modules/notifications/notifications.actions.ts
TacosTonight: Moved the filtering of active messages from notifications.selectors.ts to notifications.reducer.ts

src/client/components/common/Banner/index.tsx
```diff
@@ -0,0 +1,10 @@
+import { FC } from 'react';
+import styled from 'styled-components';
+
+interface BannerProps {
+  children: React.ReactNode;
+}
+
+export const Banner: FC<BannerProps> = ({ children }) => {
```
FiboApe, line 8: we can probably remove this file since it does nothing

src/client/data/notificationMessages.json
```diff
@@ -0,0 +1 @@
+[]
```
FiboApe, line 1: should we remove this file?

src/core/frameworks/redux/index.ts
```diff
@@ -12,10 +17,11 @@ export const getStore = (extraArgument?: any) => {
     theme: cloneDeep(themeInitialState),
     wallet: cloneDeep(walletInitialState),
     settings: cloneDeep(settingsInitialState),
+    notifications: cloneDeep(notificationsInitialState),
   };
   const persistConfig = {
     namespace: 'yearn',
-    states: ['theme', 'wallet.name', 'settings', 'network'],
+    states: ['theme', 'wallet.name', 'settings', 'network', 'notifications.dismissedMessageIds'],
```
FiboApe, line 24: this is going to make redux-localstorage-simple only load dismissedMessageIds into the store and not the other attributes from that reducer, you need to handle this by adding a preloadedState.

Take a look at https://github.com/kilkelly/redux-localstorage-simple#advanced-usage

src/core/store/modules/notifications/notifications.actions.ts
FiboApe: this will change once we know where we get the notifications from right?

src/core/store/modules/notifications/notifications.reducer.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { initialStatus, NotificationsState } from '@types';
+
+import { NotificationsActions } from './notifications.actions';
+
+export const notificationsInitialState: NotificationsState = {
+  messages: [],
+  dismissedMessageIds: [],
+  statusMap: {
+    getNotificationMessages: { ...initialStatus },
```
FiboApe, line 11: you can just do `getNotificationMessages: initialStatus`

src/core/store/modules/notifications/notifications.reducer.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    .addCase(getNotificationMessages.rejected, (state, { error }) => {
+      state.statusMap.getNotificationMessages = { loading: false, error: error.message };
+    })
+
+    .addCase(getNotificationMessages.fulfilled, (state, { payload: messages }) => {
+      state.messages = messages.filter((message) => {
+        return message.active === true;
+      });
+      state.statusMap.getNotificationMessages = {};
```
FiboApe, line 38: this should be `initialStatus` now right?

src/core/store/modules/notifications/notifications.selectors.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { RootState, Message } from '@types';
+
+/* ---------------------------------- State --------------------------------- */
+const selectDismissedMessageIds = (state: RootState) => state.notifications.dismissedMessageIds;
+const selectMessages = (state: RootState) => state.notifications.messages;
+const selectActiveMessages = createSelector<RootState, Message[], Number[], Message[]>(
+  [selectMessages, selectDismissedMessageIds],
+  (messages, dismissedMessageIds) => {
+    const activeMessages = messages.filter((message) => {
```
FiboApe, line 11: nit `messages.filter(({ id }) => !dismissedMessageIds.includes(id)`

src/core/store/modules/notifications/notifications.reducer.ts
TacosTonight: it does make sense to set it as initialStatus, since after the request is fulfilled, loading should be false and there shouldn't be an error. 

I noticed in other reducers however that the .fulfilled is set to {}

src/core/frameworks/redux/index.ts
TacosTonight: @xgambitox in the same file in like 56 there's a property inside configureStore `preloadedState: merge(initialState, persistedState)` does this handle the preloaded state?

TacosTonight:
Instead of having the Banner span the either width of the page, I tried moving the NotificationBanner inside <Content> in containers/Layout. It looks ok to me, and makes it so the fixed positioning of the side nav bar css doesn't have to be changed. The scrolling and dismissing issues that I brought up in the PR description also gets solved. What does everyone else think? @xgambitox @FiboApe 

image

src/core/store/modules/notifications/notifications.actions.ts
xgambitox: Yes, we need a decentralized source to fetch them from

src/client/data/notificationMessages.json
xgambitox: Lets leave it since we better have the importing from this file already implemented.

src/core/frameworks/redux/index.ts
xgambitox: @TacosTonight correct, that should do.

src/core/store/modules/notifications/notifications.reducer.ts
xgambitox: I prefer being more explicit and having `initialStatus ` too, but as you have it also works now since we dont have expressions like `if(getNotificationMessages.loading === false) ...`. But we need to be careful of that.

xgambitox:
> Instead of having the Banner span the either width of the page, I tried moving the NotificationBanner inside in containers/Layout. It looks ok to me, and makes it so the fixed positioning of the side nav bar css doesn't have to be changed. The scrolling and dismissing issues that I brought up in the PR description also gets solved. What does everyone else think? @xgambitox @FiboApe
> 
> image

Yeah, looks good to me :)

src/client/components/common/Banner/index.tsx
xgambitox: This component should be the presentational component. I suggest moving all the styling from `NotificationBanner` to this one. `Banner` should receive props like children and onClose.

turtlemoji:
The thing having it there as an static thing, is that will cause ugly rerenders / resizing when you close it @TacosTonight @xgambitox 
Can't we put it as position absolute/fixed at the top?

FiboApe:
So putting as position absolute is doable, but we'll have the same problem @TacosTonight was having before, where we need to push the navbars fixed position in the first render, and when the user scrolls down we need to revert that push, since the navbar will be pushed down but the notification banner not visible anymore.

As for position fixed, the issue there is that the notificationBanner will be above information that the user might want to see while scrolling through the page.

FiboApe:
Another option for this, is to have our main wrapper have an overflow hidden on the main wrapper of the app, and then on the body have an overflow scroll.

That will make it so the page always occupies the 100% of the screen, but the content can be scrolled.

So like this:
Screenshot 2022-02-02 at 11 09 08

src/client/components/common/Banner/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+export const Banner: FC<BannerProps> = ({ notification, onClose, children }) => {
+  if (!notification && !children) {
+    return null;
+  }
+
+  if (!notification) {
+    return <StyledBanner>{children}</StyledBanner>;
+  }
+
+  let messageSymbol: React.FunctionComponent<React.SVGProps<SVGSVGElement> & { title?: string | undefined }>;
```
FiboApe, line 105: you could probably use InfoIcon | WarningFilledIcon | WarningIcon | InfoIcon instead of redefining  ti as React.FC

src/client/components/common/Banner/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  display: flex;
+  align-items: center;
+  > * {
+    padding-right: 5px;
+  }
+`;
+
+const StyledBanner = styled.div`
+  background: ${({ theme }) => theme.colors.surface};
+  height: rem;
```
FiboApe, line 43: this does not actually have a value

src/client/components/common/Banner/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  position: relative;
+  padding: 30px 0px 30px 0px;
```
FiboApe, line 49: we should use rem here

src/client/components/common/Banner/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+`;
+
+const StyledNotificationBanner = styled.div`
+  height: rem;
```
FiboApe, line 53: this does not have a value

src/client/containers/NotificationBanner/index.tsx
```diff
@@ -0,0 +1,19 @@
+import { FC } from 'react';
+import { first } from 'lodash';
+
+import { useAppSelector, useAppDispatch } from '@hooks';
+import { NotificationsActions, NotificationsSelectors } from '@store';
+import { Banner } from '@components/common';
+
+export const NotificationBanner: FC = () => {
+  const latestMessage = first(useAppSelector(NotificationsSelectors.selectActiveMessages));
```
FiboApe, line 9: is there a reason we need a container component instead of just having this all on the banner component?

turtlemoji:
Refactor style, fixes, layout, animation.

PR #423, opened 2022-01-04
fix: yvBoost balance displayed correctly [WEB-1080]

gotzenx:
Since underlyingToken of yvBoost is yveCrv we were displaying the underlyingTokenBalance which in the case of "normal" vaults is correct, but since yvBoost is different we need to show vaultBalance.

As fix for labs we will display the userBalance instead of underlyingUserBalance.


turtlemoji (approved):

turtlemoji (commented):
Is pickle working as expected too?

xgambitox (approved):

PR #438, opened 2022-01-13
 feat: add health status route

xgambitox:
* Add `/health` route to check web app health status

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco start",
     "dev": "cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco start",
-    "build": "craco build",
+    "build": "cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco build",
     "test": "craco test",
     "eject": "react-scripts eject",
     "upgrade-sdk": "yarn add yearn/yearn-sdk",
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "1.0.7",
   "private": true,
   "scripts": {
-    "start": "craco start",
-    "dev": "craco start",
+    "start": "cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco start",
+    "dev": "cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco start",
     "build": "craco build",
     "test": "craco test",
     "eject": "react-scripts eject",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "bnc-notify": "1.9.1",
     "bnc-onboard": "1.34.2",
     "craco-alias": "2.3.1",
+    "cross-env": "^7.0.3",
```
turtlemoji, line 48: Remove caret

turtlemoji (dismissed):

package.json
xgambitox: true, thanks for reminder

package.json
```diff
@@ -4,8 +4,8 @@
   "private": true,
   "scripts": {
     "start": "craco start",
-    "dev": "craco start",
-    "build": "craco build",
+    "dev": "cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco start",
+    "build": "cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco build",
```
FiboApe, line 8: what is this needed for?

src/client/routes/index.tsx
```diff
@@ -49,16 +50,22 @@ const routesMap = [
...
-          {routesMap.map((route, index) => (
-            <Route key={index} exact path={route.path} component={route.component} />
-          ))}
-          <Route path="*">
-            <Redirect to="/home" />
-          </Route>
-        </Switch>
-      </Layout>
+      <Switch>
+        <Route exact path="/health" component={Health} />
```
FiboApe, line 54: is there a reason why health cannot be on the routesMap?

yarn.lock
```diff
@@ -6663,7 +6670,7 @@ cross-fetch@^2.1.0:
     node-fetch "2.6.1"
     whatwg-fetch "2.0.4"
 
-cross-spawn@7.0.3, cross-spawn@^7.0.0, cross-spawn@^7.0.2:
+cross-spawn@7.0.3, cross-spawn@^7.0.0, cross-spawn@^7.0.1, cross-spawn@^7.0.2:
```
FiboApe, line 6673: you removed the caret but need to regenerate the lockfile :)

package.json
xgambitox: Its used to have an env variable with the latest commit hash, so we know what exact version of the app is currently deployed.

src/client/routes/index.tsx
xgambitox: A health route should contain minimal content as possible so it doesnt take much server resources since we might be consuming this route frequently in our monitor system. Did not include it in the routesMap since it renders element inside the Layout component.

yarn.lock
xgambitox: Since this is a subdependency from 'cross-env', it depends on whether its pinned on there repo. Is this correct @turtlemoji ?

Did `yarn install` again but yarn.lock remains as is.

gotzenx (approved):

PR #441, opened 2022-01-14
[WEB-1161-441] feat: withdraw all from vault

xgambitox:
* Fix the insufficient funds error displayed on withdraw modal when clicking max balance
* Add withdraw all action to leave no dust on users max balance withdrawals from vaults

### DIFF
--- a/src/client/components/app/Transactions/WithdrawTx.tsx
+++ b/src/client/components/app/Transactions/WithdrawTx.tsx
@@ -154,6 +154,11 @@ export const WithdrawTx: FC<WithdrawTxProps> = ({ header, onClose, children, ...
     decimals: selectedVault.token.decimals.toString(),
     pricePerShare: selectedVault.pricePerShare,
   });
+  const percentageToWithdraw = toBN(amount)
+    .div(toUnit(underlyingTokenBalance, selectedVault.token.decimals))
+    .times(100)
+    .toString();
+  const willWithdrawAll = toBN(percentageToWithdraw).gte(99);
 
   const selectedVaultOption = {
     address: selectedVault.address,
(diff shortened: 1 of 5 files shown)


src/utils/calculations.ts
```diff
@@ -14,7 +13,22 @@ interface CalculateSharesAmountProps {
...
   if (amount.isNaN()) return '0';
-  const sharePrice = formatUnits(pricePerShare, decimals);
+  const sharePrice = toUnit(pricePerShare, parseInt(decimals));
   const ONE_UNIT = toBN('10').pow(decimals);
-  return amount.dividedBy(sharePrice).multipliedBy(ONE_UNIT).toFixed(0);
+  return amount.div(sharePrice).times(ONE_UNIT).toFixed(0);
+}
+
+interface CalculateUnderlyingAmountProps {
+  decimals: string;
```
gotzenx, line 22: would go a bit more descriptive with the props names like: yvDecimals or underlyingDecimals and so on to avoid confusion on what to send on each param.

src/client/components/app/Transactions/WithdrawTx.tsx
```diff
@@ -140,21 +142,32 @@ export const WithdrawTx: FC<WithdrawTxProps> = ({ header, onClose, children, ...
...
+  const underlyingTokenBalance = calculateUnderlyingAmount({
+    amount: toUnit(selectedVault.DEPOSIT.userBalance, parseInt(selectedVault.decimals)),
+    decimals: selectedVault.token.decimals.toString(),
+    pricePerShare: selectedVault.pricePerShare,
+  });
+  const percentageToWithdraw = toBN(amount)
+    .div(toUnit(underlyingTokenBalance, selectedVault.token.decimals))
+    .times(100)
+    .toString();
+  const willWithdrawAll = toBN(percentageToWithdraw).gte(99);
```
FiboApe, line 161: why don't we check for 100 directly?

src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -343,6 +343,67 @@ const withdrawVault = createAsyncThunk<
   }
 );
 
+const withdrawAllVault = createAsyncThunk<
+  void,
+  { vaultAddress: string; targetTokenAddress: string; slippageTolerance?: number },
+  ThunkAPI
+>(
+  'vaults/withdrawAllVault',
```
FiboApe, line 351: if you set this as `vaults/withdrawVault` then there is no need to have a different reducer cases :)

FiboApe (commented):
What's the need to have a different action for withdrawAll? Seems like its calling the same underlying service functions

src/client/components/app/Transactions/WithdrawTx.tsx
xgambitox: Price Per Share can slightly change in some cases, which leads to users having dust on their vault tokens.

src/core/store/modules/vaults/vaults.actions.ts
xgambitox: Will refactor to a new approach of reusing the withdraw action

src/utils/calculations.ts
xgambitox: done

gotzenx (approved):

turtlemoji (approved):

FoxTheSin:
QA Test: 
let simulation run after selecting "max" on the withdraw page for mainnet Link Vault.
used testing wallet to withdraw "max" SPELL, no dust remains after withdraw.

PR #450, opened 2022-01-21
[WEB-1223-450] refactor(yearnsdk): use only one instance at a time of the yearn sdk

FiboApe:
There is no actual need for us to have several instances of the Yearn sdk running at all times, and we need this to fix the  https://github.com/yearn/yearn-sdk/pull/191

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@types/redux-logger": "3.0.9",
     "@types/styled-components": "5.1.12",
     "@types/styled-system": "5.1.13",
-    "@yfi/sdk": "1.0.25",
+    "@yfi/sdk": "1.0.26",
     "awilix": "4.3.4",
     "axios": "0.21.1",
     "bignumber.js": "9.0.1",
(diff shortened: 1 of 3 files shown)


gotzenx (dismissed):

turtlemoji (approved):

turtlemoji:
We should discuss this, since the current code enables you to get and interact with all the networks at the same time, and maybe we will need to roll back later

xgambitox:
We should find a solution in the SDK that allows integrators to have multiple instances of sdk. Right now we are not currently fetching from networks at the same time, but its not necessarily true for all use cases. 

Also integrators should have the ability to use sdk allowlist validations, and at the same time being able to use their provider for other contract methods without being restricted. IMO we should change the approach of how we are trying to solve in the SDK.

FiboApe:
@xgambitox @turtlemoji agree with both, we have bounced some ideas into how to achieve this with @jstashh , this change is only a quickfix if we want to realease the allowlist now

FiboApe:
Right now we don't have the need to have multiple instances of the sdk, that will not be true forever

gotzenx:
PR will be closed since we are working on some changes on the sdk that will allow as to have multiple instances of sdk running.

PR #464, opened 2022-01-28
[WEB-1280-464] WIP: yearn.finance IPFS deployment

B3RX8OIPDHDT3:
Switches deployment to Github Actions, which includes IPFS build and publish, as well as DNS changes. Fixes https://github.com/yearn/yearn-finance-v3/issues/446.

By design, will be deployable from PRs, as well as develop+master branches: 
PR builds -> `pr-{number}.yearn.dev`.
Develop push -> `yearn.dev`.
Master push -> `yearn.finance`.

Sample PR available here, which highlights deployment steps and output: https://github.com/B3RX8OIPDHDT3/yearn-finance-v3/pull/2
Sample build available from here: https://pr-2.b3tools.xyz

**Issues/considerations:**
1. Currently IPFS build is being pushed to Pinata Cloud, is that the preferred platform for IPFS builds?
2. This workflow requires DNS updates to point to the latest IPFS cid hash. I've used Route 53 to test out DNS changes(which yearn uses I believe), and it requires two records - CNAME pointing to a gateway, as well as TXT record for DNSlink. Route 53 can set CNAME for subdomains, but it does not allow setting CNAME for a domain as per DNS spec, which means it wouldn't be able to deploy deploy(yearn.dev) and master(yearn.finance).

There are several workarounds for this, probs missing some obvious ones as well:
- Using Cloudflare for DNS since it allows setting CNAME at the apex domain level, and would work well with IPFS?
- Possibly pointing record to Cloudfront, but need to investigate how exactly it would work.

3. Should API keys be in secrets when in production, would assume yes? Current values for REACT_APP_* are taken from `.env.example`, which will need changing then.
4. Cleanup job will need to be created, which would remove DNS records + IPFS pin when PR is closed and branch is removed.

### DIFF
--- a/.github/workflows/deploy-ipfs.yml
+++ b/.github/workflows/deploy-ipfs.yml
@@ -85,6 +85,14 @@ jobs:
           aws route53 change-resource-record-sets --cli-input-json "$CHANGE_SET_CNAME"
           aws route53 change-resource-record-sets --cli-input-json "$CHANGE_SET_DNSLINK"
 
+      - name: Get Cloudflare SSL
+        run: |
+          while [ CERT_OUTPUT -ne "" ]
+          do
+            sleep 5s
+            CERT_OUTPUT=$(curl 'https://cloudflare-ipfs.com/add_host' -H 'content-type: application/x-www-form-urlencoded' --data "hostname=${{ env.DOMAIN_NAME }}" -sS | jq .err)
+          done
+
       - name: Post details to PR
         uses: actions/github-script@v5
         if: github.event_name == 'pull_request'
--- a/.github/workflows/deploy-ipfs.yml
+++ b/.github/workflows/deploy-ipfs.yml
@@ -1,0 +1,104 @@
+name: Publish to IPFS
+
+on:
+  # push:
+  #   branches:
+  #     - master
+  #     - develop
+  pull_request:
+    branches:
+      # - master
+      - develop
+
+env:
+  NODE_VERSION: '14.x'
+  IPFS_GATEWAY: 'www.cloudflare-ipfs.com.'
+  # neeeds to be set, as build currently fails because of warnings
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-ipfs.yml
```diff
@@ -0,0 +1,112 @@
...
+env:
+  NODE_VERSION: '14.x'
+  IPFS_GATEWAY: 'www.cloudflare-ipfs.com.'
+  # neeeds to be set, as build currently fails because of warnings
+  CI: false
+  AWS_HOSTED_ZONE_ID: 'Z02908602B6KLAQBRUBST'
+  AWS_DEFAULT_REGION: 'us-east-1'
+  DOMAIN_NAME: "${{ endsWith(github.ref, '/master') && 'yearn.finance' || 'b3tools.xyz' }}"
+  REACT_APP_ETHERSCAN_API_KEY: 'GEQXZDY67RZ4QHNU1A57QVPNDV3RP1RYH4'
+  REACT_APP_ALCHEMY_API_KEY: 'euSwyu6Yf-VQ3NJ32KHxDhHmTta7OvIe'
```
gotzenx, line 22: keys being public?

B3RX8OIPDHDT3: These were taken from .env.example, so pretty much public already. Just have to integrate environments into the pipeline as well, and then could have different secrets for each env.

gotzenx: Couldnt we use the variables from .env file directly? Do we need to have this here on public file duplicated?

turtlemoji:
@FoxTheSin can we get the infra team to review this?

willianpaixao:
Hi @gotzenx, the author of this PR has deleted their account. Can I pick up the task?

PR #468, opened 2022-02-01
[WEB-1300-468] feat: add error handling for simulations

FiboApe:


src/core/store/modules/vaults/vaults.reducer.ts
```diff
@@ -279,7 +279,7 @@ const vaultsReducer = createReducer(vaultsInitialState, (builder) => {
       state.statusMap.getExpectedTransactionOutcome = {};
     })
     .addCase(getExpectedTransactionOutcome.rejected, (state, { error }) => {
-      state.statusMap.getExpectedTransactionOutcome = { error: 'Simulation Failed' };
+      state.statusMap.getExpectedTransactionOutcome = { error: error.message };
```
xgambitox, line 282: wdyt if we use `parseError(error)` here in the reducer on the `rejected` actions, instead of wrapping the thrown error on the thunks?

This way we would have the whole thunk covered and no need for using try catch.

src/utils/errors.ts
```diff
@@ -0,0 +1,63 @@
+import { EthersError, PriceFetchingError, SdkError, SimulationError, TenderlyError, ZapperError } from '@yfi/sdk';
```
xgambitox, line 1: Lets create a `types/Error.ts` and re-export these error from there. We can later continue adding new types of errors from that single source.

src/utils/errors.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+const PRICE_FETCHING_ERRORS = {
+  [PriceFetchingError.FETCHING_PRICE_PICKLE]: 'Error fetching price from pickle',
+  [PriceFetchingError.FETCHING_PRICE_ORACLE]: 'Error fetching price from oracle',
+};
+const SDK_ERRORS = {
+  [SdkError.NO_SLIPPAGE]: 'No slippage set',
+};
+const SIMULATION_ERRORS = {
+  [SimulationError.NO_LOG]: 'No log of transfering token',
```
xgambitox, line 17: This error is not so clear to me. Can we add some more context or rephrase it?

src/utils/errors.ts
```diff
@@ -0,0 +1,63 @@
...
+  [SimulationError.PARTIAL_REVERT]: 'Partial rever simulating call',
```
xgambitox, line 18: Typo on `rever`

src/utils/errors.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  if (e instanceof SdkError) {
+    return (e.error_code && SDK_ERRORS[e.error_code]) || e.message;
+  }
+
+  if (e instanceof Error) {
+    return e.message;
+  }
+
+  return 'There was an error';
```
xgambitox, line 62: maybe rename to 'An Unknown Error Occurred' or similar?

src/core/store/modules/vaults/vaults.reducer.ts
gotzenx: ++
I was about to comment the same exact thing

turtlemoji: ++

src/utils/errors.ts
turtlemoji: can we translate them? There should be an errors file or object in the translates

turtlemoji: nvm, doesn't make sense to translate some errors, and have others not translated

FiboApe: I mean it does make sense to translate some errors, it's just the start of it, we can start having translation for some and then at some point have translations for them all. I can see how to consume the i18n module from here if it's alright

src/core/store/modules/vaults/vaults.reducer.ts
xgambitox: I think we have two options:
1) We adapt sdk error structure to the SerializedError structure (Seems only error_code would be required to change)
```
export interface SerializedError {
  name?: string // error_type => name ?
  message?: string
  stack?: string
  code?: string  // error_code => code
}
```
2) We add a custom `serializeError()` function to each thunk that needs it: https://redux-toolkit.js.org/api/createAsyncThunk#options

src/utils/errors.ts
turtlemoji: ok, It is just weird to sometimes display errors translated, and sometimes not

src/utils/errors.ts
FiboApe: btw this are not types, but the actual classes

src/core/store/modules/vaults/vaults.reducer.ts
xgambitox: Intention is not to couple it to a dependency, but making current naming convention more close to a standard, since the SerializedError interface is somewhat more standard than the one the sdk is currently using. Im not suggesting replacing the one we have on sdk with this one, just adapting the names basically. Think that change shouldnt avoid us from having the typescript benefits. Sdk errors can still have more properties as needed.

About wrapping the `createAsyncThunk` function, yes it can be done, and can be used accordingly depending on the use case. Its just the way the library allow us.

Im more inclined to option 1 though. Lets see what @gotzenx and @turtlemoji think.

xgambitox: Will approve since the custom serializeError is a good solution imo. We can see later if we decide to refactor :)

xgambitox (dismissed):

xgambitox (approved):

xgambitox (approved):

turtlemoji (approved):

PR #469, opened 2022-02-01
[WEB-1301-469] fix: update bnc-onboard to 1.37.2

brendanww:
We are releasing an update to Coinbase Wallet clients on Monday that is backwards incompatible with older walletlink versions for switchEthereumChain+addEthereumChain functionality. bnc-onboard 1.37.2 has walletlink 2.4.4 which the new CB Wallet clients will work correctly with.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "axios": "0.21.1",
     "bignumber.js": "9.0.1",
     "bnc-notify": "1.9.1",
-    "bnc-onboard": "1.34.2",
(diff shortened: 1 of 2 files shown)


xgambitox:
There seems to be an issue with this versions of Onboard, I could not open the wallet select modal when trying to connect wallet, then also scrolling gets stuck, at least with while testing with safari on ios.

turtlemoji:
I'm fixing the errors on the Onboard repository itself, after that we will be able to update. Closing the PR for now.

PR #472, opened 2022-02-03
feat: improved sync after tx confirmations

gotzenx:


xgambitox (approved):

gotzenx:
deprecated branch because of new approach

PR #485, opened 2022-02-13
[WEB-1354] fix: allow withdraw all from ETH

FiboApe:


src/client/components/app/Transactions/WithdrawTx.tsx
```diff
@@ -26,6 +26,7 @@ import {
 import { getConfig } from '@config';
 
 import { Transaction } from './Transaction';
+import BigNumber from 'bignumber.js';
```
turtlemoji: Move to the top

turtlemoji (dismissed):
Lgtm, would like @xgambitox to check it out to maybe abstract this logic to other place

PR #492, opened 2022-02-21
feat: validate tx using allow list [WEB-1211]

gotzenx:
Will use the sdk service `allowList` to validate the tx that are managed by the frontend since we dont depend 100% on sdk to make every tx.


FiboApe (approved):

src/core/services/TransactionService.ts
```diff
@@ -65,7 +65,17 @@ export class TransactionServiceImpl implements TransactionService {
...
 
-      // TODO call contract verification here.
+      const yearn = this.yearnSdk.getInstanceOf(network);
+      if (yearn.services.allowList) {
+        const { success: isValid, error } = await yearn.services.allowList.validateCalldata(
+          contractAddress,
+          unsignedTx.data
+        );
+        if (!isValid) {
+          if (!error) throw new Error('something went wrong but no error was provided');
```
turtlemoji, line 75: Not sure about this error, we could specify more, like "Is not valid with the Allowlist" or whatever, approved tho

turtlemoji (approved):

src/core/services/TransactionService.ts
```diff
@@ -65,7 +65,17 @@ export class TransactionServiceImpl implements TransactionService {
       // const decodedData = contractIface.decodeFunctionData(methodName, unsignedTx.data!.toString());
       // console.log({ decodedData });
```
xgambitox, line 66: can we remove this commented code now?

src/core/services/TransactionService.ts
gotzenx: But we can NOT know if tx is not valid or something exploded on the validation method. Thats why we should not assume that if `isValid` returned false is because tx is not valid. SDK returns isValid = false if there was any kind of error on the method. For example, if the SC call fails due to network acting weird, sdk returns isValid = false, but there was no actual validation. The call to the SC just expoded for some random reason.

src/core/services/TransactionService.ts
gotzenx: At first i did removed it but then i thought maybe it was useful to have it there just in case of need. It could be very possible that we will need to use it on `TransactionService` at some point

xgambitox (dismissed):

turtlemoji (dismissed):

FiboApe (dismissed):

FoxTheSin:
SDK allow list has been updated (https://github.com/yearn/yearn-sdk/pull/220)

FiboApe (approved):

xgambitox (approved):

Issue #497, opened 2022-02-23
fix: tw.card img [WEB-1371]


FoxTheSin:
**Describe the bug**
When yearn.finance/#/vaults is linked (e.g. in twitter if you link the site) the thumbnail icon isn't showing up

**To Reproduce**
1. go to twitter
2. post link to yearn.finance/#/vaults

**Expected behavior**
A thumbnail of yearn should show up on all links

**Screenshots**
Displays the blank "gray paper" instead of a nice thumbnail (like notion)

![Screen Shot 2022-02-23 at 8 51 43 AM](https://user-images.githubusercontent.com/79029523/155367107-fe8735d9-4f9a-476b-aab8-17540a8c96ee.png)

**Environment Information**

**Additional context**

karelianpie:
Fixed with https://github.com/yearn/yearn-finance-v3/pull/503

PR #503, opened 2022-02-26
fix: tw.card img

karelianpie:
Fixes https://github.com/yearn/yearn-finance-v3/issues/497

It appears that `%PUBLIC_URL%` is not working in the `content` tag, only `href` (like for the manifest file) so I've changed it to be static. It should now allow twitter to grab this image https://yearn.finance/share.png for the thumbnail

yearn_finance


turtlemoji (approved):
Tested and working, thanks!

xgambitox (approved):

Issue #493, opened 2022-02-22
fix: yeild under modal displaying 0


FoxTheSin:
**Elevator Pitch**
When a new vault is created or introduced, no historical pricing is available. The Modal that presents "yield" % will show 0, should be listed as NEW (reflecting the vaults page)

**Customer Value**
Wont scare people with a 0% apy

**Definition of Done**
when website is loaded, the yield will display NEW (with the little star emoji)

**Design Approach**
Approach and considerations to build the task list / Existing code patterns in production code base that you will base your work off of.

**Additional Context**
Screenshots, UI/UX designs, links or other that will help implementing the feature.
![image](https://user-images.githubusercontent.com/79029523/155201204-374cf3c2-9929-48d1-8398-9ee80ff76457.png)

karelianpie:
@FoxTheSin I was looking into this but unsure where is the modal that you're referring to, could you please let me know

xgambitox:
In the particular case of yvBoost, it is due to an error in the APY calcs. We can display a hyphen '-' to distinguish those vaults with `apyMetadata.type === 'error'` so we can distinguish them.

karelianpie:
> In the particular case of yvBoost, it is due to an error in the APY calcs. We can display a hyphen '-' on those vaults with `apyMetadata.type === 'error'` so we can distinguish them.

Sounds good to me. I'll add that to the `formatApy` util

PR #504, opened 2022-02-26
fix: Yield under modal displaying 0

karelianpie:
Fixes https://github.com/yearn/yearn-finance-v3/issues/493

Uses the util function and finds the type (e.g "new") from the apy metadata

### DIFF
--- a/src/client/components/app/Transactions/DepositTx.tsx
+++ b/src/client/components/app/Transactions/DepositTx.tsx
@@ -21,7 +21,7 @@ import {
   validateSlippage,
   validateNetwork,
   getZapInContractAddress,
-  formatPercent,
+  formatApy,
 } from '@utils';
 import { getConfig } from '@config';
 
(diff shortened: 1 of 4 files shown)


karelianpie:
It appears that the team is working on a solution to fix the 0.00% in the case of yvBOOST vault (https://t.me/c/1152055703/21767), the fix in this current PR will work for vaults that have `apyType === "new"`

gotzenx:
This fix should also be implemented on `DepositTx` component to impact the vaults deposits modal and not only the labs deposits modal. Right?

xgambitox:
as @gotzenx says, lets add the fix to `DepositTx` component too

karelianpie:
@gotzenx @xgambitox added the `formatApy` to `DepositTx` with https://github.com/yearn/yearn-finance-v3/pull/504/commits/26427afde6ce2e686b54af6ea425e434ad57bf54

Made `apyType` optional so we don't need to pass it as metadata is optional with https://github.com/yearn/yearn-finance-v3/pull/504/commits/ac1d18cfc07f85cea6c90c08031b14c49e17e72a

xgambitox (approved):

PR #511, opened 2022-03-04
fix: tx success exit button on vault details [WEB-851]

turtlemoji:
The message in vault details after a successful tx should be "Done" and the button work as expected.


src/client/components/app/Transactions/Transaction.tsx
```diff
@@ -37,7 +37,8 @@ interface Asset {
 interface TransactionProps {
   transactionLabel?: string;
   transactionCompleted: boolean;
-  transactionCompletedLabel: string;
+  setTxCompleted?: Dispatch<SetStateAction<boolean>>;
```
turtlemoji, line 40: This is my only concern...

turtlemoji: 1) Type. it is an oddly specific one I know.
2) @xgambitox do you know if we need to clear some state too between txs? Because I saw some TODOS you left about clearing the state in some cases

xgambitox:
I would have used a different approach:

- Use the `onClose` prop to differentiate when the component is rendered from a modal or from vaults detail view.
- Also change the `transactionCompletedLabel` text based on the above.
- Use `onTransactionCompletedDismissed` to handle the action of either executing `onClose` or reset the components state back to original with `setTxCompleted(false)`.

turtlemoji:
> 

I thought about it, but in this case, that would case to have more duplicated code across every TxContainer, like, DepositTx, WithdrawTx, MigrateTx, ClaimTx, etc.

I would stop treating the transaction component as a dumb component, and start migrating logic there. Since it is indeed a generic but smart component that handles lot of stuff, and it is the same stuff for every txcontainer.

wdyt about dumb vs smart in this case?

src/client/components/app/Transactions/DepositTx.tsx
```diff
@@ -26,6 +26,7 @@ import {
 import { getConfig } from '@config';
 
 import { Transaction } from './Transaction';
+import { TxStatusActionType } from './components/TxStatus';
```
xgambitox, line 29: No longer need this import right?

src/client/components/app/Transactions/MigrateTx.tsx
```diff
@@ -12,6 +12,7 @@ import {
 } from '@utils';
 
 import { Transaction } from './Transaction';
+import { TxStatusActionType } from './components/TxStatus';
```
xgambitox, line 15: Import not needed anymore?

src/client/components/app/Transactions/Transaction.tsx
```diff
@@ -37,7 +37,8 @@ interface Asset {
 interface TransactionProps {
   transactionLabel?: string;
   transactionCompleted: boolean;
-  transactionCompletedLabel: string;
+  transactionCompletedLabel?: string;
+  transactionCompletedActionType?: TxStatusActionType;
```
xgambitox, line 41: can remove this too?

xgambitox (approved):

gotzenx (approved):

PR #513, opened 2022-03-07
Integrate the Rave Names FNS into the wallet address display

Pokesi:
Uses the `fns-helper` module to allow for display of Rave Names alongside ENS.


FoxTheSin:
> Uses the `fns-helper` module to allow for display of Rave Names alongside ENS.

Thanks for reaching out Pokesi.
In a bid to minimize external dependencies, we'd like everything coded in our repository. If you can make this happen, great, please start a new PR as we'll be closing this one.

PR #514, opened 2022-03-07
feat: add partner id integration

FiboApe:


src/core/frameworks/yearnSdk/index.ts
```diff
@@ -1,20 +1,22 @@
...
+    const { SUPPORTED_NETWORKS, CONTRACT_ADDRESSES } = config;
 
+    const isLedger = inLedgerIframe();
     SUPPORTED_NETWORKS.forEach((network) => {
       const providerType = getProviderType(network);
       const provider = web3Provider.getInstanceOf(providerType);
       const networkId = getNetworkId(network) as SdkNetwork;
       const sdkInstance = new Yearn(networkId, {
         provider,
+        partnerId: isLedger ? CONTRACT_ADDRESSES.LEDGER_PARTNER_ID : '',
```
xgambitox, line 19: Can we set 'partnerId' to 'undefined' better?

xgambitox (dismissed):

turtlemoji (dismissed):

FoxTheSin:
amplify not working?
site isn't loading https://pr-514.d27dgpz01hmbvx.amplifyapp.com/

src/core/store/modules/tokens/tokens.reducer.ts
```diff
@@ -147,6 +151,27 @@ const tokensReducer = createReducer(tokensInitialState, (builder) => {
...
+        [spenderAddress]: allowance,
+      };
+      state.statusMap.user.userTokensActionsMap[tokenAddress].getAllowances = {};
+      state.statusMap.user.getUserTokensAllowances = {};
+    })
+    .addCase(getVaultAllowance.rejected, (state, { meta, error }) => {
+      const { tokenAddress } = meta.arg;
+      state.statusMap.user.userTokensActionsMap[tokenAddress].getAllowances = { error: error.message };
+      state.statusMap.user.getUserTokensAllowances = { error: error.message };
+    })
```
xgambitox, line 174: should we move all this logic to the vaults reducer instead?

src/core/store/modules/tokens/tokens.reducer.ts
```diff
@@ -160,6 +185,13 @@ const tokensReducer = createReducer(tokensInitialState, (builder) => {
         ...state.user.userTokensAllowancesMap[tokenAddress],
         [spenderAddress]: amount,
       };
+    })
+    .addCase(approveDeposit.fulfilled, (state, { meta, payload: { amount, spenderAddress } }) => {
+      const { tokenAddress } = meta.arg;
+      state.user.userTokensAllowancesMap[tokenAddress] = {
+        ...state.user.userTokensAllowancesMap[tokenAddress],
+        [spenderAddress]: amount,
+      };
```
xgambitox, line 194: should we move all this logic to the vaults reducer instead?

src/core/types/Service.ts
```diff
@@ -161,6 +165,28 @@ export interface GetTokenAllowanceProps {
...
+  accountAddress: Address;
+  amount: Wei;
+  vault: Vault;
+}
+
+export interface GetVaultAllowanceProps {
+  network: Network;
+  accountAddress: string;
+  tokenAddress: string;
+  vault: { address: string; token: string };
```
xgambitox, line 187: We should try to use a flat object when possible and not nested objects.

src/client/components/app/Transactions/DepositTx.tsx
```diff
@@ -67,6 +72,10 @@ export const DepositTx: FC<DepositTxProps> = ({
     : userTokens;
   const sellTokensOptionsMap = keyBy(sellTokensOptions, 'address');
   const selectedSellToken = sellTokensOptionsMap[selectedSellTokenAddress ?? ''];
+  const [tokenAllowance, isLoadingTokenAllowance, getTokenAllowanceErrors] = useAllowance(
```
xgambitox, line 75:
```suggestion
  const [tokenAllowance, isLoadingTokenAllowance, tokenAllowanceError] = useAllowance(
```
Lets remove the `get` since it makes it look like if its a function

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
```diff
@@ -64,6 +68,10 @@ export const LabDepositTx: FC<LabDepositTxProps> = ({ onClose }) => {
     : userTokens;
   const sellTokensOptionsMap = keyBy(sellTokensOptions, 'address');
   const selectedSellToken = sellTokensOptionsMap[selectedSellTokenAddress ?? ''];
+  const [tokenAllowance, isLoadingTokenAllowance, getTokenAllowanceErrors] = useAllowance(
```
xgambitox, line 71:
```suggestion
  const [tokenAllowance, isLoadingTokenAllowance, tokenAllowanceError] = useAllowance(
```
Lets remove the `get` since it makes it look like if its a function

src/client/hooks/useAllowance.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AppServiceContext } from '@context';
+import { NetworkSelectors, WalletSelectors } from '@store';
+import { TokenAllowance } from '@types';
+import { useAppSelector } from '@hooks';
+
+import { usePrevious } from './usePrevious';
+
+export function useAllowance(
+  tokenAddress: string | undefined,
+  vault: { address: string; token: { address: string } } | undefined
```
xgambitox, line 13: We should try to use a flat object when possible and not nested objects.

src/client/context/AppServiceContext.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export const AppServiceContext = createContext<ServiceContainer | undefined>(undefined);
+
+interface AppServiceContextProviderProps {
+  services: ServiceContainer;
+}
+
+export const AppServiceContextProvider: FC<AppServiceContextProviderProps> = ({ children, services }) => {
+  return <AppServiceContext.Provider value={services}>{children}</AppServiceContext.Provider>;
+};
```
xgambitox, line 13: To maintain the currents system architecture, we should avoid UI calling services directly, we use redux actions as an adapter layer. As an alternative, we can dispatch a thunk and we can unwrap the value, with no reducer needed.

src/client/hooks/useAllowance.ts
```diff
@@ -0,0 +1,56 @@
...
+    async function callPromise() {
+      if (selectedAddress && vault && tokenAddress && walletIsConnected) {
+        try {
+          const promiseResult = await services?.vaultService.getVaultAllowance({
+            network: currentNetwork,
+            vault: { address: vault.address, token: vault.token.address },
+            tokenAddress,
+            accountAddress: selectedAddress,
+          });
+          console.log('fetched allowance', promiseResult);
```
xgambitox, line 35: remove console log

src/client/context/AppServiceContext.tsx
xgambitox: Since we are out of time for this feature, we can leave it for now and aim to refactor it immediately after release.

src/core/frameworks/yearnSdk/index.ts
```diff
@@ -1,20 +1,25 @@
...
     SUPPORTED_NETWORKS.forEach((network) => {
       const providerType = getProviderType(network);
       const provider = web3Provider.getInstanceOf(providerType);
       const networkId = getNetworkId(network) as SdkNetwork;
       const sdkInstance = new Yearn(networkId, {
         provider,
+        partnerId: isLedger ? CONTRACT_ADDRESSES.LEDGER_PARTNER_ID : undefined,
+        cache: {
+          useCache: false,
+        },
```
xgambitox, line 22: can remove

src/core/types/Service.ts
FiboApe: sure will change that

src/client/components/app/Transactions/DepositTx.tsx
FiboApe: will do

src/client/context/AppServiceContext.tsx
FiboApe: what's the logic behind not allowing hooks or ui not call services directly? It seems that just adding a redux action as an intermediary is overhead just to keep a pattern that might not make sense in all cases.

It seems like we would just use redux for the thunk action but actually not do anything with the result from it

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
FiboApe: solved on 6ad01db108237880b69ecb49695a74b419be2063

src/client/components/app/Transactions/DepositTx.tsx
FiboApe: solved on 6ad01db108237880b69ecb49695a74b419be2063

src/client/hooks/useAllowance.ts
FiboApe: solved on fff3e86f6179d4e466b091c8f555551d66f809ef

src/core/types/Service.ts
FiboApe: solved on fff3e86f6179d4e466b091c8f555551d66f809ef

src/client/hooks/useAllowance.ts
FiboApe: solved on 305142a804e3c9fe05a74cb42aecce53b2d45f4e

src/core/frameworks/yearnSdk/index.ts
FiboApe: solved on ecc064b992e17714c48485b4293dee5e8c5f2590

src/client/context/AppServiceContext.tsx
xgambitox: We are applying the [Clean Architecture](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html) principles. We are trying to maintain the business logic independent of the UI. Maintaining these separation between layers, each having their own responsibilities, make code more maintainable, flexible and scalable. In this case, actions are part of our Interface Adapters layer.

If you take a look at how it was implemented before this change, it was following this pattern. Although maintaining this architecture can be seen as having too much boilerplate, we designed it this way so that we get all the benefits from it, including utilities from the tooling itself. We add validations there, since we dont want to only have them on UI side, also makes them callable by other thunks and not just directly from UI components. Right now we are missing some critical stuff that will also benefit from this structure, like error handling and analytics through middleware, and of course testing for the core logic (actions, reducers, services).

src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -174,16 +175,35 @@ const getExpectedTransactionOutcome = createAsyncThunk<
...
+      network: network.current,
+      accountAddress,
+      tokenAddress,
+      amount,
+      vault: vaultData,
+    });
+
+    if (typeof tx !== 'boolean') {
+      await transactionService.handleTransaction({ tx: tx as TransactionResponse, network: network.current });
+    }
```
xgambitox, line 204: Why these changes? Is it even called now that the service is called directly?

src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -192,12 +212,25 @@ const approveDeposit = createAsyncThunk<void, { vaultAddress: string; tokenAddre
...
+    const tx = await vaultService.approveZapOut({
+      network: network.current,
+      accountAddress,
+      amount,
+      vault: vaultData,
+    });
+
+    if (typeof tx !== 'boolean') {
+      await transactionService.handleTransaction({ tx: tx as TransactionResponse, network: network.current });
+    }
```
xgambitox, line 233: same comment [here](https://github.com/yearn/yearn-finance-v3/pull/514/files#r838029524)  for this changes

src/client/hooks/useAllowance.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { AppServiceContext } from '@context';
+import { NetworkSelectors, WalletSelectors } from '@store';
+import { TokenAllowance } from '@types';
+import { useAppSelector } from '@hooks';
+
+import { usePrevious } from './usePrevious';
+
+export function useAllowance(
+  tokenAddress: string | undefined,
```
karelianpie, line 12:
```suggestion
  tokenAddress?: string,
```

src/client/hooks/useAllowance.ts
```diff
@@ -0,0 +1,60 @@
...
+import isEqual from 'lodash/isEqual';
+
+import { AppServiceContext } from '@context';
+import { NetworkSelectors, WalletSelectors } from '@store';
+import { TokenAllowance } from '@types';
+import { useAppSelector } from '@hooks';
+
+import { usePrevious } from './usePrevious';
+
+export function useAllowance(
```
karelianpie, line 11: Can we have here named params?

FiboApe: solved on ffefab04728c259b147ce88165c34d609bd1c7b4

src/client/context/AppServiceContext.tsx
FiboApe: solved on ffefab04728c259b147ce88165c34d609bd1c7b4

src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -174,16 +176,35 @@ const getExpectedTransactionOutcome = createAsyncThunk<
...
+  async ({ vaultAddress, tokenAddress }, { dispatch, getState, extra }) => {
+    const { vaults, wallet, network } = getState();
     const vaultData = vaults.vaultsMap[vaultAddress];
-    const isZapin = vaultData.tokenId !== tokenAddress;
-    const spenderAddress = isZapin ? getConfig().CONTRACT_ADDRESSES.zapIn : vaultAddress;
+    const { vaultService, transactionService } = extra.services;
+    const amount = extra.config.MAX_UINT256;
+
+    const accountAddress = wallet.selectedAddress;
+    const isZapin = vaultData.token === tokenAddress;
```
gotzenx, line 192: should this be `!==`?

gotzenx (dismissed):

src/core/store/modules/vaults/vaults.actions.ts
FiboApe: fixed with 020ad2bba3da6fa711b571e9e3cbee69c0e3b331

src/core/types/Service.ts
```diff
@@ -55,6 +56,9 @@ export interface VaultService {
   deposit: (props: DepositProps) => Promise<TransactionResponse>;
   withdraw: (props: WithdrawProps) => Promise<TransactionResponse>;
   migrate: (props: MigrateProps) => Promise<TransactionResponse>;
+  approveDeposit: (props: ApproveDepositProps) => Promise<TransactionResponse | Boolean>;
+  approveZapOut: (props: ApproveZapOutProps) => Promise<TransactionResponse | Boolean>;
```
xgambitox, line 60: With new refactor on sdk, we can remove the `Boolean` from return type
```suggestion
  approveDeposit: (props: ApproveDepositProps) => Promise<TransactionResponse>;
  approveZapOut: (props: ApproveZapOutProps) => Promise<TransactionResponse>;
```

src/core/services/VaultService.ts
```diff
@@ -256,4 +260,30 @@ export class VaultServiceImpl implements VaultService {
...
+    const yearn = this.yearnSdk.getInstanceOf(network);
+
+    return yearn.vaults.approveDeposit(accountAddress, vaultAddress, tokenAddress, amount);
+  }
+
+  public async approveZapOut(props: ApproveZapOutProps): Promise<Boolean | TransactionResponse> {
+    const { network, vault, accountAddress } = props;
+    const yearn = this.yearnSdk.getInstanceOf(network);
+
+    return yearn.vaults.approveWithdraw(vault.address, vault.token, accountAddress);
```
xgambitox, line 275: the Token argument should be the intended target token to withdraw to. Its not always the underlying token (vault.token). Would suggest we just receive directly the addresses instead of the whole Vault data structure.

src/core/types/Service.ts
karelianpie: Also, we should always use lowercase `boolean` as that's the primitive type (not `Boolean`).

src/core/services/VaultService.ts
```diff
@@ -256,4 +260,30 @@ export class VaultServiceImpl implements VaultService {
         });
     }
   }
+
+  public async approveDeposit(props: ApproveDepositProps): Promise<Boolean | TransactionResponse> {
```
karelianpie, line 264:
```suggestion
  public async approveDeposit(props: ApproveDepositProps): Promise<boolean | TransactionResponse> {
```

src/core/types/Service.ts
xgambitox: Yes, that was an error on the SDK that was removed in the recent version, so Fibo probably was forced to copy that interface. Agree we should stick with primitive. Also we should not have this type of interfaces that can return multiple types.

src/core/services/VaultService.ts
FiboApe: fixed on 227dce8ccc7427637245d1ffbe9f21f0819cece8

src/core/services/VaultService.ts
FiboApe: fixed on 227dce8ccc7427637245d1ffbe9f21f0819cece8

src/core/types/Service.ts
FiboApe: fixed on 227dce8ccc7427637245d1ffbe9f21f0819cece8

src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -174,30 +176,62 @@ const getExpectedTransactionOutcome = createAsyncThunk<
...
+
+    const tx = await vaultService.approveDeposit({
+      network: network.current,
+      accountAddress,
+      tokenAddress,
+      amount,
+      vaultAddress: vaultData.address,
+    });
+
+    if (typeof tx !== 'boolean') {
```
xgambitox, line 203: Can remove this now that TX type changed

src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -174,30 +176,62 @@ const getExpectedTransactionOutcome = createAsyncThunk<
...
+      await transactionService.handleTransaction({ tx: tx as TransactionResponse, network: network.current });
+    }
 
-    const result = await dispatch(TokensActions.approve({ tokenAddress, spenderAddress }));
-    unwrapResult(result);
+    return { amount, spenderAddress: isZapin ? getConfig().CONTRACT_ADDRESSES.zapIn : vaultAddress };
```
xgambitox, line 207: Since spender can be also partner tracking contract, would be best to dispatch `getVaultAllowance` after transaction is approved, so that state gets updated correctly ?

FiboApe: fixed on b69e516c61c60f31501bd291946417d970f1b2ad

src/core/store/modules/vaults/vaults.actions.ts
FiboApe: fixed on b69e516c61c60f31501bd291946417d970f1b2ad

xgambitox (approved):

FoxTheSin:
updated testing may have come across new error with approval
image_2022-04-07_15-21-26
image_2022-04-07_15-17-58

xgambitox (approved):

FoxTheSin:
errors on approve appear resolved. 
tested fantom deposit look good
tested eth deposit simulations look good

FoxTheSin (approved):

PR #515, opened 2022-03-07
feat: integrate the Rave Names FNS into the wallet address display.

Pokesi:
This time using just ethers :)

Just allows for Rave resolution alongside ENS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "eslint-config-prettier": "8.3.0",
     "eslint-plugin-prettier": "3.4.0",
     "ethers": "5.4.5",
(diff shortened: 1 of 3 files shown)


FoxTheSin:
Sorry pokesi, haven't had time to prioritize reviewing this.

Key side note, please make sure this doesn't create a dependency in our code (for security purposes we are minimizing any dependencies we use.)

Pokesi:
> Key side note, please make sure this doesn't create a dependency in our code (for security purposes we are minimizing any dependencies we use.)

Yeah, it uses ethers.js now :)

PR #525, opened 2022-03-10
Add wallet is connected before getTokenAllowance

### LINKED ISSUE #507: fix: vault details page approve button
**Describe the bug**
When user loads the website directly on Vaults Detail View, fetching of token allowance is being made before the wallet is connect. This results in the fetch throwing an error so the app never gets the user allowance correctly. This is the reason why we will always display the `Approve` token button as enabled since logic doesnt have the allowance value.

**To Reproduce**
1. Identify a vault that you know you have approve the underlying token to deposit.
2. Navigate to that vaults detail view by clicking on it from the vaults list.
3. Once inside the Detail View, refresh the web
Note: on redux dev tools you will find that the action `getTokenAllowance` got rejected with error: WALLET NOT CONNECTED.
You should use the dev build that has redux logging enabled: https://develop.d27dgpz01hmbvx.amplifyapp.com/#/vaults

**Expected behavior**
- We should fetch `getTokenAllowance` successfully after wallet is connected. 
- Approve button should be displayed as disabled and `deposit` button enabled.

**Screenshots**
image
image
Note: on the second image we can see that we are dispatching getTokenAllowace before walletChange action

**Environment Information**

**Additional context**

phillip055:
fixes #507

### DIFF
--- a/src/core/store/modules/wallet/wallet.selectors.ts
+++ b/src/core/store/modules/wallet/wallet.selectors.ts
@@ -2,7 +2,7 @@ import { getNetwork } from '@utils';
 import { RootState } from '@types';
 
 const selectWallet = (state: RootState) => state.wallet;
-const selectWalletIsConnected = (state: RootState) => state.wallet.isConnected;
+const selectWalletIsConnected = (state: RootState) => state.wallet.isConnected && !!state.wallet.selectedAddress;
 const selectSelectedAddress = (state: RootState) => state.wallet.selectedAddress;
 const selectAddressEnsName = (state: RootState) => state.wallet.addressEnsName;
 const selectWalletNetwork = (state: RootState) =>
--- a/src/core/store/modules/wallet/wallet.selectors.ts
+++ b/src/core/store/modules/wallet/wallet.selectors.ts
@@ -2,7 +2,6 @@ import { getNetwork } from '@utils';
 import { RootState } from '@types';
 
 const selectWallet = (state: RootState) => state.wallet;
-const selectWalletIsLoading = (state: RootState) => state.wallet.isLoading;
 const selectWalletIsConnected = (state: RootState) => state.wallet.isConnected;
 const selectSelectedAddress = (state: RootState) => state.wallet.selectedAddress;
 const selectAddressEnsName = (state: RootState) => state.wallet.addressEnsName;
@@ -11,7 +10,6 @@ const selectWalletNetwork = (state: RootState) =>
 
 export const WalletSelectors = {
   selectWallet,
-  selectWalletIsLoading,
   selectWalletIsConnected,
   selectSelectedAddress,
   selectAddressEnsName,
--- a/src/core/store/modules/wallet/wallet.selectors.ts
+++ b/src/core/store/modules/wallet/wallet.selectors.ts
@@ -2,7 +2,8 @@ import { getNetwork } from '@utils';
 import { RootState } from '@types';
 
 const selectWallet = (state: RootState) => state.wallet;
-const selectWalletIsConnected = (state: RootState) => !!state.wallet.selectedAddress;
+const selectWalletIsLoading = (state: RootState) => state.wallet.isLoading;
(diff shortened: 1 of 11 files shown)


turtlemoji:
@phillip055 not resolved for me, in vault details, the approve button is enabled when it should be disabled.
We are still not dispatching the correct action to get the user allowance.

turtlemoji (changes requested):
issue is still there after the changes

src/core/store/modules/wallet/wallet.selectors.ts
```diff
@@ -2,14 +2,16 @@ import { getNetwork } from '@utils';
 import { RootState } from '@types';
 
 const selectWallet = (state: RootState) => state.wallet;
-const selectWalletIsConnected = (state: RootState) => !!state.wallet.selectedAddress;
+const selectWalletIsLoading = (state: RootState) => state.wallet.isLoading;
```
turtlemoji, line 5: remove

src/core/store/modules/wallet/wallet.selectors.ts
```diff
@@ -2,14 +2,16 @@ import { getNetwork } from '@utils';
...
+const selectWalletIsConnected = (state: RootState) => state.wallet.isConnected;
 const selectSelectedAddress = (state: RootState) => state.wallet.selectedAddress;
 const selectAddressEnsName = (state: RootState) => state.wallet.addressEnsName;
 const selectWalletNetwork = (state: RootState) =>
   state.wallet.networkVersion ? getNetwork(state.wallet.networkVersion) : undefined;
 
 export const WalletSelectors = {
   selectWallet,
+  selectWalletIsLoading,
```
turtlemoji, line 14: idem

turtlemoji:
Working fine now!

turtlemoji:
@phillip055 can you add this logic to the migrateTx component too?
And I think we will need it too on "LockTx", "ReinvestTx", "LabDepositTx", "LabStakeTx", "LabWithdrawTx" since all of them have allowance, even if they are not in the vault details and will not need the fix soon, it will be nice to have it already implemented.

@xgambitox wdyt?

turtlemoji (dismissed):

src/core/store/modules/wallet/wallet.selectors.ts
```diff
@@ -2,7 +2,7 @@ import { getNetwork } from '@utils';
 import { RootState } from '@types';
 
 const selectWallet = (state: RootState) => state.wallet;
-const selectWalletIsConnected = (state: RootState) => !!state.wallet.selectedAddress;
+const selectWalletIsConnected = (state: RootState) => state.wallet.isConnected;
```
xgambitox, line 5: I remember we changed this a while ago since sometimes onboard remains connected but no longer have a selected address. There where some edge cases where that happened, like when selecting another wallet and then dismissing the prompt. Would suggest we dont change this unless its really necessary, and have someone test it thoroughly.

xgambitox: Change was done here, while testing multi chain support, https://github.com/yearn/yearn-finance-v3/commit/7bec8af03e4e7056bc52653581d013b5c3531aac#diff-031294351bc496adc7c6604ac7cdc16dc2a287110e5a14931d9c3e0a7edcbdc2L4

turtlemoji: should we do `const selectWalletIsConnected = (state: RootState) => state.wallet.isConnected && !!state.wallet.selectedAddress` 

@xgambitox wdyt

turtlemoji (approved):

src/core/store/modules/wallet/wallet.selectors.ts
turtlemoji: @xgambitox pushed the logic from the above comment, tested and working fine, but we should have fox test too

xgambitox (approved):

PR #534, opened 2022-03-13
fix: toggle switch on/off clarity [DES-144]

pentcle:
use success colour in each theme for toggled-on switch background

use grey + nearly-white for off state

image

image

image


FoxTheSin:
Testing looks great <3

turtlemoji (approved):
After discussion with pentacle, we decided to leave the refactor to revamp after merging this, approved!

xgambitox (approved):

PR #540, opened 2022-03-15
Ambire Wallet batch transaction implementation through WalletConnect

0xdex18:
Allowing Ambire Wallet users to bundle approval transaction with other approval required actions on yearn

### DIFF
--- a/src/client/components/app/Transactions/DepositTx.tsx
+++ b/src/client/components/app/Transactions/DepositTx.tsx
@@ -24,6 +24,7 @@ import {
   formatApy,
 } from '@utils';
 import { getConfig } from '@config';
+import useIsAmbireWC from '@hooks/useIsAmbireWC';
 
 import { Transaction } from './Transaction';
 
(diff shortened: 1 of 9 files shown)


turtlemoji:
Hey @0xdex18 , can't you get in contact with onboard js library to implement the wallet?
So we don't have custom logic on the FE, wdyt?

Ivshti:
hey @turtlemoji - this PR does not implement Ambire

In fact, Ambire is already accessible through WalletConnect

However, it enables automatic batching of approval+action thanks to Ambire's ability to batch transactions, which is impossible to shim at a provider level, therefore not able to be done in onboard

turtlemoji:
> hey @turtlemoji - this PR does not implement Ambire
> 
> In fact, Ambire is already accessible through WalletConnect
> 
> However, it enables automatic batching of approval+action thanks to Ambire's ability to batch transactions, which is impossible to shim at a provider level, therefore not able to be done in onboard

K! Have a quick look. I'm not really sure about having this custom logic inside vault actions.
Also you are only implementing logic on `withdraw, suppy, deposit` and not other modals, is this correct/intended?.

I think @xgambitox will have a better insight here

Issue #530, opened 2022-03-11
fix: exit button for "curve lock" modal


FoxTheSin:
**Describe the bug**
for the lab yveCRV lock, the modal that pops up doesn't have an exit button

**To Reproduce**
When you are on ethereum network for yearn.finance

1. Go to Labs

2. Click on Lock (only on yveCRV vault)

3. No exit button

**Expected behavior**
There should be an exit button (like image below)

![image](https://user-images.githubusercontent.com/79029523/157926726-9c764227-eb97-4ed1-96f8-e99208e17c86.png)

**Screenshots**
What it does look like

![image](https://user-images.githubusercontent.com/79029523/157926847-c8f979ae-cd8a-43c3-9cbf-2cee514e37f8.png)

kyhoon:
i had a try on this issue at #545

PR #545, opened 2022-03-16
fix: pass the onClose prop to ModalTx

kyhoon:
Suggestion for #530

assuming that the `onClose` function could be called either when
- the modal is closed
- `onTransactionCompletedDismissed` is called from the transaction

I have added the onClose prop to the wrapper modal as well. Let me know if this is not the intended behavior.

And also I would like to note that a number of modals (e.g. Backscractchers) share this same structure, so if we could work with this I may add this to the other TxModals as well

### DIFF
--- a/src/client/components/common/TxModal.tsx
+++ b/src/client/components/common/TxModal.tsx
@@ -10,8 +10,14 @@ const StyledModalTx = styled(Modal)`
   height: initial;
 `;
 
-export interface StyledModalTxProps {}
+export interface StyledModalTxProps {
+  onClose?: () => void;
+}
 
-export const ModalTx: FC<StyledModalTxProps> = ({ children, ...props }) => {
-  return <StyledModalTx {...props}>{children}</StyledModalTx>;
+export const ModalTx: FC<StyledModalTxProps> = ({ onClose, children, ...props }) => {
+  return (
+    <StyledModalTx onClose={onClose} {...props}>
+      {children}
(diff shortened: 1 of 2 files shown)


xgambitox:
Hey @kyhoon! yes that is the intended behavior, solution is good. If you fix other modals that might have same issue, please include here in same PR. Will approve for now :)

xgambitox (approved):

kyhoon:
alright, I will try to add this to the other modals too

xgambitox:
Hey there @kyhoon , still up to finish the request?

If not we will merge soon as is and rest can be added in another PR.

kyhoon:
sorry for the delay 😓, I did make a mental note on this but did not make further changes yet.
Sounds good, I will come back and pick this up again after the merge if there is no one else that wants to handle this in the meantime

FoxTheSin:
Review, merge, and create another PR for the rest of the work please @kyhoon

turtlemoji:
@kyhoon I spotted that the current issue was that we were missing the onClose prop on the lock modal wrapper, for the transaction component. It was not needed to add it to the lockmodal wrapper itself, also this could have impacted since we were using the onClose prop to know when it was a modal, and when it was a fixed component inside the vault details, so thats why we decided to just fix the onclose prop and not to add it to the wrapper!
Can you do a quick review on the #581  PR? Ty so much!

turtlemoji:
closed for #581

PR #555, opened 2022-03-23
Testing AWS Amplify (to be removed)

karelianpie:
Testing if AWS Amplify is broken :)

To be removed


karelianpie:
Amplify was actually failing correctly, it's just not reporting to github.

PR #558, opened 2022-03-25
fix: Get metadata overrides for `yvBoost` from `vaults.getDynamic()`

### LINKED ISSUE #524: fix: vaults from the labs page do not read overrides from yearn-meta
**Describe the bug**
yvBoost apy is overriden in meta - https://github.com/yearn/yearn-meta/blob/master/data/vaults/1/0x9d409a0A012CFbA9B15F6D4B36Ac57A46966Ab9a.json

however it's not being used for displaying the apy %

Screenshot 2022-03-10 at 12 55 40

**To Reproduce**
1. Go to labs screen
2. View yvboost

**Expected behavior**
Data is read from yearn meta so that apys are overriden if necessary

**Environment Information**
 - Web/Brave

karelianpie:
Closes https://github.com/yearn/yearn-finance-v3/issues/524

Get metadata overrides for `yvBoost` from `vaults.getDynamic()`, the `displayIcon` is overwritten in order to be the yvToken icon itself.

yearn_finance

### DIFF
--- a/src/core/services/LabService.ts
+++ b/src/core/services/LabService.ts
@@ -141,38 +141,47 @@ export class LabServiceImpl implements LabService {
     // **************** YVBOOST ****************
     let yvBoostLab: Lab | undefined;
     try {
-      const yvBoostContract = getContract(YVBOOST, yvBoostAbi, provider);
-      const totalAssets = await yvBoostContract.totalAssets();
-
-      const yvBoostVaultDynamic = (await yearn.vaults.getDynamic([YVBOOST]))[0];
+      const [yvBoostVaultDynamic] = await yearn.vaults.getDynamic([YVBOOST]);
       const yvBoostData = vaultsResponse.data.find(({ address }: { address: string }) => address === YVBOOST);
 
       // TODO We could use the data from `yvBoostVaultDynamic`
-      if (!yvBoostData) {
+      if (!yvBoostData || !yvBoostVaultDynamic) {
         throw new Error(`yvBoost vault not found on ${YEARN_API} response`);
       }
 
+      if (!yvBoostVaultDynamic) {
+        throw new Error(`yvBoost vault not found on yearn dynamic vaults response`);
+      }
+
+      const { address, decimals, name, symbol, token, version } = yvBoostData;
+      const { underlyingTokenBalance } = yvBoostVaultDynamic;
+
+      const amount = underlyingTokenBalance.amount.toString();
+
       const yveCrvPrice = pricesResponse.data['vecrv-dao-yvault']['usd'];
 
+      // This amount usdc is slightly different from the one that comes from `underlyingTokenBalance`
+      const amountUsdc = toBN(normalizeAmount(amount, decimals))
+        .times(yveCrvPrice)
+        .times(10 ** USDC_DECIMALS)
+        .toFixed(0);
+
       yvBoostLab = {
-        address: YVBOOST,
+        address,
         typeId: 'LAB',
-        token: YVECRV,
-        name: yvBoostData.name,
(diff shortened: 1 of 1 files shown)


xgambitox (approved):

src/core/services/LabService.ts
```diff
@@ -148,6 +149,7 @@ export class LabServiceImpl implements LabService {
         yvBoostContract.emergencyShutdown(),
       ]);
       const yvBoostData = vaultsResponse.data.find(({ address }: { address: string }) => address === YVBOOST);
+      const { apyOverride } = await yearn.tokens.findMetadataOverrides(CONTRACT_ADDRESSES.YVBOOST);
```
karelianpie, line 152: ⚠️ We don't have this yet, it's in https://github.com/yearn/yearn-sdk/pull/263

karelianpie:
@xgambitox this PR was using the function `findMetadataOverrides` from https://github.com/yearn/yearn-sdk/pull/263 but that was removed with https://github.com/yearn/yearn-sdk/pull/263/commits/69fa49237a012eb0028d8fb6e5f968853aef1670 since we've thought that a better approach would be to get it from `vaults.getDynamic()` instead.

Refactored with https://github.com/yearn/yearn-finance-v3/pull/558/commits/6d0d73b6b8da9553f86135017e3b44f907177219

src/core/services/LabService.ts
karelianpie: N/A

src/core/services/LabService.ts
```diff
@@ -141,15 +142,18 @@ export class LabServiceImpl implements LabService {
     let yvBoostLab: Lab | undefined;
     try {
       const yvBoostContract = getContract(YVBOOST, yvBoostAbi, provider);
-      const [totalAssets, pricePerShare, depositLimit, emergencyShutdown] = await Promise.all([
-        yvBoostContract.totalAssets(),
-        yvBoostContract.pricePerShare(),
-        yvBoostContract.depositLimit(),
-        yvBoostContract.emergencyShutdown(),
-      ]);
+      const totalAssets = await yvBoostContract.totalAssets();
```
xgambitox, line 145: I think we can also get this value from `yearn.vaults.getDynamic` result. Try `yvBoostVaultDynamic.underlyingTokenBalance.amount` if its the same as `totalAssets`. If that is correct, then also use the `amountUsdc` value from response instead of us continuing to calculated manually.

karelianpie: You're right, they are the same:

```typescript
totalAssets.toString()
// "12970596589987347531198138"

yvBoostVaultDynamic.underlyingTokenBalance.amount.toString()
// "12970596589987347531198138"
```

Refactored with https://github.com/yearn/yearn-finance-v3/pull/558/commits/0dc62b3fcc4ee2f9272fc733645c17f74d8d11d8 to used getDynamic

xgambitox: Yes, that `pricesResponse` comes from coingecko and `yvBoostVaultDynamic` from oracle contract, so should be more real, unless there is low liquidity. If prices are similar, lets go with the later, but we need to make sure price is consistent here too https://github.com/yearn/yearn-finance-v3/blob/develop/src/core/services/TokenService.ts#L116 if we decide to change it.

karelianpie: Refactored with https://github.com/yearn/yearn-finance-v3/pull/558/commits/0f3e5cfe76f529a412a520457113b908b03e68cf to use the balances from the `underlyingTokenBalance` from the getDynamic vaults.

FiboApe (approved):
Looks good! Nice work!

xgambitox (approved):

Issue #532, opened 2022-03-12
fix: rainbow - wallet connect mobile


FoxTheSin:
**Describe the bug**
Rainbow wallet on mobile isn't playing well with yearn.finance

**To Reproduce**
1. On a mobile device (potentially on desktop as well?) connect to yearn.finance
2. Connect a Rainbow wallet
3. open a transaction modal (depsoit or withdraw)
4. doesn't work

**Expected behavior**
Should be able to conduct transactions with rainbow wallet on mobile

**Screenshots**
![image](https://user-images.githubusercontent.com/79029523/158037760-a333592d-99d2-4ea6-9c52-cc61d39c213e.png)

**Environment Information**

**Additional context**

pentcle:
steps taken having cleared safari cache on ios


go to yearn.finance

![image0](https://user-images.githubusercontent.com/80323528/158038452-ac8f9abb-04db-462b-b49f-49f46fc57dd4.png)


tap to connect wallet

![image1](https://user-images.githubusercontent.com/80323528/158038451-d013f7cd-fa6c-46ce-a04e-97ea3e337a5f.png)


tap rainbow

![image2](https://user-images.githubusercontent.com/80323528/158038449-55688fb9-115e-4a67-8175-2fdab2c9f1e2.png)


tap to connect

![image3](https://user-images.githubusercontent.com/80323528/158038448-e8cf54e8-2339-447e-9ab1-de76bf44a04a.png)


why am i now at the app store? 

![image4](https://user-images.githubusercontent.com/80323528/158038446-2f147dc3-97a0-459a-9d84-d8ea157f85a1.png)


open rainbow wallet app on phone (was hazarding a guess to try that)

![image5](https://user-images.githubusercontent.com/80323528/158038444-493ac316-dce2-430d-974f-4de52307f7e1.png)


return to yearn.finance, my wallet is connected

![image0](https://user-images.githubusercontent.com/80323528/158038569-ab101b77-bbca-4f4b-bb4a-b408bcb5b6c2.png)

pentcle:
step ii - am now connected via rainbow wallet

tap to deposit ETH
 
![image1](https://user-images.githubusercontent.com/80323528/158039027-15af5121-06e3-45ea-9e40-2b8e4d367eec.png)

ignore warning message and tap to deposit max

[...]

PR #560, opened 2022-03-25
fix: Convert `networkId` to string

karelianpie:
Fixes https://github.com/yearn/yearn-finance-v3/issues/532

The wallet network id for Rainbow with WalletConnect comes as a `string`, so in `src/utils/network.ts` in the `getNetwork()` we return "other" rather than "mainnet", even though the network is `"1"`, but "1" as string and not as a number. By parsing the networkId as a number we don't get the warning anymore.

**Update:** Refactored to convert the `networkId` to String instead following this discussion https://github.com/yearn/yearn-finance-v3/pull/560#discussion_r835659000

telegram-cloud-photo-size-5-6195093049051295172-y

With MetaMask we get a number

telegram-cloud-document-5-6195093048595055853


src/utils/network.ts
```diff
@@ -23,7 +23,7 @@ export const getNetworkId = (network: Network): number => {
 };
 
 export const getNetwork = (networkId: number): Network => {
-  switch (networkId) {
+  switch (Number(networkId)) {
```
turtlemoji, line 26: Maybe will be better to use toString for more error redundancy but lgtm

turtlemoji (dismissed):

src/utils/network.ts
karelianpie: Sounds good, refactored with https://github.com/yearn/yearn-finance-v3/pull/560/commits/ee17a64d506836ddac93678e0b4187200251ef00

karelianpie: I was wondering if we should throw or log the error somewhere, it wasn't the most obvious bug to find. Do you know if we are allowing for network to be "other" and if we do anything with it or would we be better off throwing and logging an error.

```typescript
throw Error(`Unknown Network Id: ${networkId}`);
```

It would have helped with finding this bug

karelianpie: Also added `string` to type of `networkId` with https://github.com/yearn/yearn-finance-v3/pull/560/commits/3e8cc2a808d9c6735936eff2a7099d29de99f081

turtlemoji: I think we can throw an error too yeah sounds smart, and display an alert to change the network to the current selected network in the drop-down. But maybe that alert should be handled in other place and PR.

karelianpie: Agreed, refactored with https://github.com/yearn/yearn-finance-v3/pull/560/commits/99a3ceb760022ca043fc19c9ae76c5d19594dce9 (added a few tests as well)

turtlemoji (dismissed):

turtlemoji (approved):
lgtm!, lets wait until @xgambitox can confirm that we don't need the 'other' network in any other place (there is a type were it is included)

karelianpie:
> lgtm!, lets wait until @xgambitox can confirm that we don't need the 'other' network in any other place (there is a type were it is included)

We can also remove from the `Network` type if we don't make use of it. Thoughts @xgambitox

xgambitox:
Users should be able to set the wallet to an unsupported network and app should continue to work properly, just wont be able to execute a transaction. The `wallet.networkVersion` state of the store can be set to any networkId, since it gets set to users current network, not just supported ones. The problem with throwing an error there is that it breaks in some places if you are in an unsupported chain. You can try browsing the app with `BSC` or `Polygon` and try to open the deposit modal to see app breaks.

xgambitox (changes requested):
Users should be able to set the wallet to an unsupported network and app should continue to work properly, just wont be able to execute a transaction. The `wallet.networkVersion` state of the store can be set to any networkId, since it gets set to users current network, not just supported ones. The problem with throwing an error there is that it breaks in some places if you are in an unsupported chain. You can try browsing the app with `BSC` or `Polygon` and try to open the deposit modal to see app breaks.

karelianpie:
> Users should be able to set the wallet to an unsupported network and app should continue to work properly, just wont be able to execute a transaction. The `wallet.networkVersion` state of the store can be set to any networkId, since it gets set to users current network, not just supported ones. The problem with throwing an error there is that it breaks in some places if you are in an unsupported chain. You can try browsing the app with `BSC` or `Polygon` and try to open the deposit modal to see app breaks.

That makes sense, refactored with https://github.com/yearn/yearn-finance-v3/pull/560/commits/95823368eb9d3c8e6b3955743826d64930912245 to continue returning 'other' but log warn so we can more easily debug these network issues. Thoughts @xgambitox

turtlemoji (approved):

xgambitox (approved):

Issue #538, opened 2022-03-14
feat: vault details page for all Lab vaults


FoxTheSin:
**Elevator Pitch**
We need a vault details page specifically for labs.

**Customer Value**
Customers would be able to see the contract link and get a better idea of what the heck veCRV , yvboost, and teh pickle jar is all about.

**Definition of Done**
When clicked either in the "Highest APY" card or "opportunities" card, should link to the Vault Details Page

**Design Approach**
Approach and considerations to build the task list / Existing code patterns in production code base that you will base your work off of.

**Additional Context**
![image](https://user-images.githubusercontent.com/79029523/158248565-f980f768-1688-4e78-be4c-e9bc20271447.png)

![image](https://user-images.githubusercontent.com/79029523/158248639-c2ab0792-d1e1-455d-a294-2195cd65fa54.png)

DarkGhost7:
This is helpful because at times I go to the website to find the vault contract by click on the name of the vault, I believe if it doesnt have a vault detail page I cannot find the vault contract for that vault from the website.

karelianpie:
Created a PR for this issue, https://github.com/yearn/yearn-finance-v3/pull/561

PR #561, opened 2022-03-26
feat: Add Lab details page

karelianpie:
Closes https://github.com/yearn/yearn-finance-v3/issues/538

Started adding a Lab details page for all the labs' tokens, however it seems that only the yvBOOST lab has a description and unsure if it is correct. Actions and translations were not added, let's add those in a different PR.

yearn_finance

You can now click on both the top level card and the list

yearn_finance


public/locales/en/labdetails.json
```diff
@@ -0,0 +1,25 @@
+{
```
karelianpie, line 1: Some of these might not be necessary for labs, if anyone with a little bit more domain knowledge knows which aren't needed please let me know and I'll remove them.

src/client/components/app/LabDetail/LabDetailPanels.tsx
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  return (
+    <Tooltip placement="bottom" tooltipComponent={<ApyTooltipData apy={apyMetadata} address={address} />}>
+      <StyledIcon Component={InfoIcon} size="1.5rem" />
+    </Tooltip>
+  );
+};
+
+export const LabDetailPanels = ({
```
karelianpie, line 197: This and the `LabDetail` share a lot in common with their Vault counterpart, do you think it's worth to DRY so they share the same code? I wasn't sure if the cleaning up was worth it if we are not planning in adding more of these master/detail screens

turtlemoji:
I would check if it is better to just modify the vault details component instead of the duplicate as @karelianpie said first

karelianpie:
> I would check if it is better to just modify the vault details component instead of the duplicate as @karelianpie said first

Was looking into decoupling for a couple of hours and it's not straightforward. Firstly, the routes map in `src/client/routes/index.tsx` is giving the specific details page component to the `react-router-dom`'s `Route`, then inside each detail page we have specific hooks to fetch data depending on the data source, the same in the detail panels. So these would be tricky to decouple and in my opinion not worth it at this point, if we were to introduce extra data sources maybe it would be worth it to put effort into refactoring, but just for two master/detail views I don't think it's necessary and more time efficient to support the current code. Let me know your thoughts @turtlemoji

xgambitox:
Yeah, might not be worth now I agree. Also, Labs might have different and particular behavior from regular vaults, so decoupling them makes sense. Maintaining them both is more work, but allows us to add custom stuff. We should aim later to just make the  components used in both more composable if necessary.

turtlemoji:
Yeah, the best thing then would be to have some shared components then. I really don't like to have it all duplicated as I am afraid we will leave it like that forever since it will work fine.
If not done in this PR we should do it pretty soon on another one

turtlemoji:
code lgtm, need testing

turtlemoji (approved):

DarkGhost7:
just chiming in here this description is not for yvBoost but for yveCRV vault. Also if you dont know if a strat description is right or if another one is needs ping me and ill help get it figured out

karelianpie:
> just chiming in here this description is not for yvBoost but for the yveCRV vault. Also if you dont know if a strat description is right or if another one is needs ping me and ill help get it figured out

@DarkGhost7 thanks for the heads-up, we were discussing this in the weekly today. @FoxTheSin can you please follow-up on this as we thought that even though the description is for `yveCRV` it's the one to have for this lab.

The other 2 labs are missing the descriptions.

FoxTheSin:
Thanks @DarkGhost7 ! For sure @karelianpie I believe we need to update yearn meta, since they never originally intended to have descriptions/about sections.

src/client/routes/Labs/index.tsx
```diff
@@ -481,8 +483,7 @@ export const Labs = () => {
                 </SearchBarContainer>
               }
               searching={opportunities.length > filteredOpportunities.length}
-              // TODO Redirect address is wrong
-              // onAction={({ address }) => history.push(`/vault/${address}`)}
+              onAction={({ address }) => history.push(`/lab/${address}`)}
```
xgambitox, line 486: Also need same fix [here](https://github.com/yearn/yearn-finance-v3/pull/561/files#diff-656bc72c3ac9fd94e17e141c522cd0a9c54a12dae5221f8b981f091dc44a7d65R411)

xgambitox:
Code looks good, we can refactor some functional components later with https://github.com/yearn/yearn-finance-v3/issues/570. @FoxTheSin please test with preview build and add corresponding data needed in yearn-meta. Think the old figma from designs has some copy that might be useful.

src/client/components/app/LabDetail/LabDetailPanels.tsx
```diff
@@ -0,0 +1,287 @@
...
+              </InfoValueRow>
+            </TokenInfo>
+          </OverviewTokenInfo>
+
+          <StyledCardHeader header={t('labdetails:overview-panel.about')} />
+
+          {selectedLab.token.description && (
+            <OverviewInfo variant="surface" cardSize="micro">
+              <StyledCardContent>
+                <Markdown>{selectedLab.token.description}</Markdown>
```
xgambitox, line 279: Beware places that `selectedLab.token` is used. In Vaults page the displayed data is mostly of the underlying token. In Labs we do have to display the vault data itself.

karelianpie: Does this mean we might need to get the `description` from somewhere else? `selectedLab.underlyingToken` or similar?

src/client/routes/Labs/index.tsx
karelianpie: What fix? I'm confused :)

src/client/components/app/LabDetail/LabDetailPanels.tsx
xgambitox: `token` already holds the information of the underlying token. Not sure what is the best route here. Maybe `LabView` should have additional properties as needed.

src/client/routes/Labs/index.tsx
xgambitox: You missed adding the `history.push` fix here https://github.com/yearn/yearn-finance-v3/blob/54d4240ba6ebb586850fd9bb1f9e79b5864e665d/src/client/routes/Labs/index.tsx#L411

src/client/components/app/LabDetail/LabDetailPanels.tsx
karelianpie: Shall we leave that to another PR?

src/client/routes/Labs/index.tsx
karelianpie: Ahhh fixed with https://github.com/yearn/yearn-finance-v3/pull/561/commits/ec31c34f3dc2028038661cf909f60e47058983d1

src/client/components/app/LabDetail/LabDetailPanels.tsx
xgambitox: Lets try to keep it here since releasing it without all the info, makes the use case of having this view very limited.

FoxTheSin:
I think mostly cosmetic/copy/ux/ui stuff, but the lab details show up for all 3!
so there's yv boost balance
image


but withdraw/deposit modal is missing
image

FoxTheSin:
image

FoxTheSin:
Has this warning thing always been there?

image

karelianpie:
> but withdraw/deposit modal is missing 

The `DepositTx` and `WithdrawTx` components are tightly coupled with vaults, that's why I've left them out in this first iterations. It would be nice to get the details view out and then add those actions later.

src/client/routes/LabDetail/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const walletName = useAppSelector(WalletSelectors.selectWallet);
+
+  const currentNetworkSettings = NETWORK_SETTINGS[currentNetwork];
+  const blockExplorerUrl = currentNetworkSettings.blockExplorerUrl;
+
+  const [firstTokensFetch, setFirstTokensFetch] = useState(false);
+  const [tokensInitialized, setTokensInitialized] = useState(false);
+
+  useEffect(() => {
+    const assetAddress: string | undefined = location.pathname.split('/')[2];
```
FiboApe, line 71: we should be able to use `https://v5.reactrouter.com/core/api/Hooks/useparams` useParams instead of splitting location here

src/client/routes/LabDetail/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    if (!assetAddress || !isValidAddress(assetAddress)) {
+      dispatch(AlertsActions.openAlert({ message: 'INVALID_ADDRESS', type: 'error' }));
+      history.push('/home');
```
FiboApe, line 74: What do you think about adding enums for our Routes? In case they change later on we do not need to redefine everywhere when we use it

src/client/routes/LabDetail/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    dispatch(LabsActions.setSelectedLabAddress({ labAddress: assetAddress }));
+
+    return () => {
+      dispatch(LabsActions.clearSelectedLabAndStatus());
+    };
+  }, [currentNetwork]);
+
+  useEffect(() => {
+    const loading = tokensStatus.loading;
+    if (loading && !firstTokensFetch) setFirstTokensFetch(true);
```
FiboApe, line 86: I'm having a bit of a hard time following why we need the logic from firstTokensFetch

src/client/components/app/LabDetail/LabDetailPanels.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+const StyledCardHeaderContainer = styled.div`
+  display: flex;
+  align-items: center;
+  justify-items: between;
+`;
+
+const StyledImg = styled.img`
+  object-fit: cover;
+  width: 30px;
```
FiboApe, line 42: should we use rem here?

src/client/components/app/LabDetail/LabDetailPanels.tsx
```diff
@@ -0,0 +1,330 @@
...
+  currentNetwork,
+  blockExplorerUrl,
+}: LabDetailPanelsProps) => {
+  const { t } = useAppTranslation('labdetails');
+
+  const context = useContext(AppContext);
+
+  const [selectedTab, setSelectedTab] = useState('deposit');
+
+  const handleTabChange = (selectedTab: string) => {
```
FiboApe, line 231: should we have a type describing which tabs exist?

src/client/components/app/LabDetail/LabDetailPanels.tsx
```diff
@@ -0,0 +1,330 @@
...
+  right: 0;
+  top: 0;
+  width: 50%;
+`;
+
+const OverviewInfo = styled(Card)`
+  padding: ${({ theme }) => theme.card.padding};
+  flex-shrink: 0;
+
+  a {
```
FiboApe, line 64: Where are we using a raw `a` link?

src/client/routes/LabDetail/index.tsx
karelianpie: Thanks for the review @FiboApe, all this code came from the Vault details, so unsure how much effort we want to put into refactoring it since the idea short term is to move this more to the SDK

@xgambitox what are your thoughts?

Issue #531, opened 2022-03-11
feat: Add apy to selectSUmmaryData selector for labs


FoxTheSin:
**Elevator Pitch**
Add APY for the vaults holdings section

**Customer Value**
Right now, this is the returned data for selectSummaryData selector for labs:

return {
    totalDeposits: totalDeposited.toString(),
    totalEarnings: '0',
    estYearlyYeild: '0',
  };

But we need to add apy per request of the design team as:

The holdings value is the totalDeposits, so we are only missing the APY value in the selector.

**Definition of Done**
the APY is displayed through code
![image](https://user-images.githubusercontent.com/79029523/157988089-63cca829-6f26-4b3f-b376-a248f9724470.png)

**Design Approach**
Approach and considerations to build the task list / Existing code patterns in production code base that you will base your work off of.

**Additional Context**
Screenshots, UI/UX designs, links or other that will help implementing the feature.

karelianpie:
Started looking into this one, created a draft PR https://github.com/yearn/yearn-finance-v3/pull/562

Wondering how is this APY calculated? Suppose you have 3 deposites, what would the APY be? I'm imagining that we would need to weigh each deposit/apy and then get the weighted average apy?

xgambitox:
This issue was not clear enough. In these context `APY` represents the `Estimated Yearly Yield` for the connected wallet. In the home screen, for vaults, this value is fetched from the SDK, which aggregates this data from subgraph. To replicate this for Labs, its more of a manual task, trying to emulate what the subgraph does, since subgraph does not aggregate data for this special vaults (only yvBoost).

To obtain the `estimated yearly yield`, we need to calculate the `estimated earnings` for each vault a user has holdings (calculated using current APY), then `sum of all estimated earnings in Labs` / `sum of holdings in Labs`

PR #562, opened 2022-03-27
feat: Add `totalEarnings` and `estYearlyYield` to selectSummaryData selector for labs [WEB-1406]

karelianpie:
Closes https://github.com/yearn/yearn-finance-v3/issues/531

Calculates and displays the `totalEarnings` and `estYearlyYield` in the summary data for labs.

Screenshot_30_3_2022__2_19_pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "redux-logger": "3.0.6",
     "styled-components": "5.3.0",
     "styled-system": "5.1.5",
-    "type-fest": "^2.12.1",
+    "type-fest": "2.12.1",
     "typescript": "4.3.5",
     "web-vitals": "1.1.2"
   },
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "redux-logger": "3.0.6",
     "styled-components": "5.3.0",
     "styled-system": "5.1.5",
-    "type-fest": "2.12.1",
     "typescript": "4.3.5",
     "web-vitals": "1.1.2"
   },
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "redux-logger": "3.0.6",
     "styled-components": "5.3.0",
     "styled-system": "5.1.5",
+    "type-fest": "^2.12.1",
     "typescript": "4.3.5",
     "web-vitals": "1.1.2"
   },
--- a/src/core/store/modules/labs/labs.selectors.ts
+++ b/src/core/store/modules/labs/labs.selectors.ts
@@ -165,6 +165,7 @@ const selectSummaryData = createSelector([selectDepositedLabs], (depositedLabs) 
   depositedLabs.forEach((lab) => (totalDeposited = totalDeposited.plus(lab[lab.mainPositionKey].userDepositedUsdc)));
 
   return {
+    apy: depositedLabs.map((lab) => lab.apyMetadata?.net_apy),
     totalDeposits: totalDeposited.toString(),
     totalEarnings: '0',
     estYearlyYeild: '0',
--- a/src/core/store/modules/labs/labs.selectors.ts
+++ b/src/core/store/modules/labs/labs.selectors.ts
@@ -13,7 +13,7 @@ import {
   LabsPositionsTypes,
 } from '@types';
 import { getConstants } from '@config/constants';
-import { calculateCombinedApy, toBN } from '@utils';
+import { computeSummaryData, toBN } from '@utils';
 
 import { createToken } from '../tokens/tokens.selectors';
 
@@ -161,17 +161,7 @@ const selectSelectedLab = createSelector([selectLabs, selectSelectedLabAddress],
 });
 
 const selectSummaryData = createSelector([selectDepositedLabs], (depositedLabs) => {
-  let totalDeposited = toBN('0');
-  depositedLabs.forEach((lab) => (totalDeposited = totalDeposited.plus(lab[lab.mainPositionKey].userDepositedUsdc)));
-
-  return {
-    apy: calculateCombinedApy(
-      depositedLabs.map(({ apyMetadata, DEPOSIT: { userDepositedUsdc } }) => [apyMetadata?.net_apy, userDepositedUsdc])
-    ),
-    totalDeposits: totalDeposited.toString(),
-    totalEarnings: '0',
-    estYearlyYeild: '0',
-  };
+  return computeSummaryData(depositedLabs);
 });
 
 const selectLabsStatus = createSelector(
--- a/src/core/store/modules/labs/labs.selectors.ts
+++ b/src/core/store/modules/labs/labs.selectors.ts
@@ -13,7 +13,7 @@ import {
   LabsPositionsTypes,
 } from '@types';
 import { getConstants } from '@config/constants';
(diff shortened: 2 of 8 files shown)


src/utils/calculations.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { PartialDeep } from 'type-fest';
```
xgambitox, line 1: For the sake of not adding another package, do you think Partial<T> can be enough here? Just trying to find a balance between the value it can add to have more type safety vs adding a dependency. Tree shaking should take care of dependency in the build, so it should not increase the size of the bundled js.

package.json
```diff
@@ -74,6 +74,7 @@
     "redux-logger": "3.0.6",
     "styled-components": "5.3.0",
     "styled-system": "5.1.5",
+    "type-fest": "^2.12.1",
```
xgambitox, line 77: We should pin the dependency for security reasons. (if we decide to add)

src/core/store/modules/labs/labs.selectors.ts
```diff
@@ -165,6 +165,9 @@ const selectSummaryData = createSelector([selectDepositedLabs], (depositedLabs)
   depositedLabs.forEach((lab) => (totalDeposited = totalDeposited.plus(lab[lab.mainPositionKey].userDepositedUsdc)));
 
   return {
+    apy: calculateCombinedApy(
+      depositedLabs.map(({ apyMetadata, DEPOSIT: { userDepositedUsdc } }) => [apyMetadata?.net_apy, userDepositedUsdc])
+    ),
     totalDeposits: totalDeposited.toString(),
     totalEarnings: '0',
     estYearlyYeild: '0',
```
xgambitox, line 173: Lets populate these ones, based on the additional context of the issue https://github.com/yearn/yearn-finance-v3/issues/531#issuecomment-1081321273

src/utils/calculations.spec.ts
karelianpie: `Partial` is not enough here, we can just define the `PartialDeep` ourselves, but that `type-fest` is commonly used.

package.json
karelianpie: My bad, keep forgetting it. Pinned with https://github.com/yearn/yearn-finance-v3/pull/562/commits/325ce7f07d7cd4803fa8d5a97b4c12aa24e5f39e

src/utils/calculations.spec.ts
xgambitox: I would say that if we only use this one, we can just define the type ourselves yes. If later we see we need more special types, then we can think of adding the whole lib...

@turtlemoji what do you think?

karelianpie: Using our own `DeepPartial` with https://github.com/yearn/yearn-finance-v3/pull/562/commits/59f6de9dabffff9a8f7f80b5daf3daa0209908fc, `type-fest` removed 😢

src/core/store/modules/labs/labs.selectors.ts
karelianpie: Done with https://github.com/yearn/yearn-finance-v3/pull/562/commits/35b7318b06ad491e43fe2b9b8295a38c7f7adb15

src/utils/calculations.ts
```diff
@@ -32,3 +32,22 @@ export function calculateUnderlyingAmount(props: CalculateUnderlyingAmountProps)
...
+    {
+      totalDeposits: new BigNumber(0),
+      totalEarnings: new BigNumber(0),
+    }
+  );
+
+  return {
+    totalDeposits: totalDeposits.toString(),
+    totalEarnings: totalEarnings.toString(),
+    estYearlyYield: totalDeposits.isZero() ? '0' : totalEarnings.div(totalDeposits).toString(),
```
turtlemoji, line 51: should we fix the decimals here?

src/utils/calculations.spec.ts
turtlemoji: Having the type ourselves and not having to think about prototype pollution or dependency injection in another lib is amazing for me.
If it was for me just even trim lodash too 😈  @xgambitox @karelianpie

src/utils/calculations.ts
karelianpie: In `src/client/routes/Labs/index.tsx` it formats correctly, see the screenshot in the PR description

```typescript
          { header: t('dashboard.holdings'), Component: <Amount value={totalDeposits} input="usdc" /> },
          { header: 'Earnings', Component: <Amount value={totalEarnings} input="usdc" /> },
          { header: 'Est. Yearly Yield', content: `${formatApy(estYearlyYield)}` },
```

turtlemoji: Oh yes, formatApy does the trick sorry, approved!

turtlemoji (dismissed):

src/utils/calculations.spec.ts
karelianpie: +1 for no lodash for me as well :)

xgambitox: I <3 lodash

src/utils/calculations.ts
```diff
@@ -32,3 +32,22 @@ export function calculateUnderlyingAmount(props: CalculateUnderlyingAmountProps)
...
+
+export function computeSummaryData(labs: Pick<GeneralLabView, 'apyData' | 'DEPOSIT'>[]): SummaryData {
+  const { totalDeposits, totalEarnings } = labs.reduce(
+    ({ totalDeposits, totalEarnings }, { DEPOSIT: { userDepositedUsdc }, apyData }) => ({
+      totalDeposits: totalDeposits.plus(userDepositedUsdc),
+      totalEarnings: totalEarnings.plus(toBN(userDepositedUsdc).times(apyData)),
+    }),
+    {
+      totalDeposits: new BigNumber(0),
+      totalEarnings: new BigNumber(0),
```
xgambitox, line 44:
```suggestion
      totalDeposits: toBN(0),
      totalEarnings: toBN(0),
```
We always try to use our function just so its also easier to replace in the future in case is needed.

xgambitox:
Earnings in Vaults page displays the actual amount that has been earned by using Yean vaults. That is calculated by the subgraph, and is not implemented for us to obtain same consistent data for labs right now.

Just so we dont confuse users, we should change the `Earnings` label to `Est. Yearly Earnings` on Labs, so it makes more sense with current calcs, since is the info we have available at the moment.

src/utils/calculations.ts
karelianpie: Can't believe I forgot to use that 🤦 

https://github.com/yearn/yearn-finance-v3/pull/562/commits/60264391382a0ce9cd32a144131835cd86aa87e1

karelianpie:
> Just so we dont confuse users, we should change the `Earnings` label to `Est. Yearly Earnings` on Labs, so it makes more sense with current calcs, since is the info we have available at the moment.

Yeah, was thinking the same, refactored with https://github.com/yearn/yearn-finance-v3/pull/562/commits/cb7dcfba582effbc61438ac1b7780d5fd9773f1b

xgambitox (approved):

PR #563, opened 2022-03-27
Feat/revamp UI

pentcle:
have included 'macro' on styled-component imports for debugging

begun general cleaning of markup - very minimal - eg using headings vs divs, and articles vs divs where appropriate

### DIFF
--- a/src/client/components/app/Navbar/index.tsx
+++ b/src/client/components/app/Navbar/index.tsx
@@ -1,4 +1,4 @@
-import styled from 'styled-components';
+import styled from 'styled-components/macro';
 
 import { ConnectWalletButton } from '@components/app';
 import { OptionList, EthereumIcon, FantomIcon, ArbitrumIcon } from '@components/common';
@@ -29,7 +29,7 @@ const StyledNavbarActions = styled.div`
 const StyledText = styled.h1<{ toneDown?: boolean }>`
   display: inline-flex;
   font-size: 2.4rem;
-  font-weight: bold;
+  font-weight: 700;
   color: ${({ theme }) => theme.colors.titles};
   margin: 0;
   padding: 0;
--- a/src/client/components/common/Button/index.tsx
+++ b/src/client/components/common/Button/index.tsx
@@ -1,5 +1,5 @@
 import { FC } from 'react';
-import styled from 'styled-components';
+import styled from 'styled-components/macro';
 
 export interface ButtonProps {
   className?: string;
@@ -14,7 +14,7 @@ const StyledButton = styled.button<{ outline?: boolean }>`
   flex-direction: row;
   align-items: center;
   justify-content: center;
-  padding: 0 1rem;
+  padding: 0 1.6rem;
   height: 2.8rem;
   border: 2px solid transparent;
   border-radius: ${({ theme }) => theme.globalRadius};
@@ -23,7 +23,7 @@ const StyledButton = styled.button<{ outline?: boolean }>`
   font-family: inherit;
   cursor: pointer;
   user-select: none;
-  font-size: 1.4rem;
+  font-size: 1.6rem;
   transition: filter 200ms ease-in-out;
 
   a,
--- a/src/client/components/common/Card/CardElement/index.tsx
+++ b/src/client/components/common/Card/CardElement/index.tsx
@@ -1,9 +1,9 @@
 import { FC, ReactNode } from 'react';
-import styled from 'styled-components';
+import styled from 'styled-components/macro';
 
 import { Icon, ArrowDownIcon, IconProps } from '@components/common';
 
-const Container = styled.div<{ width?: string; align?: string; grow?: string; fontWeight?: number }>`
+const Container = styled.article<{ width?: string; align?: string; grow?: string; fontWeight?: number }>`
   display: flex;
   flex-direction: column;
   width: ${({ width }) => width ?? '17rem'};
@@ -34,10 +34,13 @@ const SortIcon = styled(({ activeSort, sortType, ...props }: SortIconProps) => <
   `}
 `;
 
-const Header = styled.div<{ onClick?: () => void }>`
+const Header = styled.h3<{ onClick?: () => void }>`
   display: flex;
   align-items: center;
   font-size: 1.6rem;
+  font-weight: 400;
+  margin: 0;
+  padding: 0;
   color: ${({ theme }) => theme.colors.texts};
   cursor: ${({ onClick }) => (onClick ? 'pointer' : 'default')};
 `;
--- a/src/client/components/common/Card/CardElement/index.tsx
+++ b/src/client/components/common/Card/CardElement/index.tsx
@@ -39,14 +39,13 @@ const Header = styled.h3<{ onClick?: () => void }>`
   align-items: center;
   font-size: 1.6rem;
   font-weight: 400;
-  margin: 0;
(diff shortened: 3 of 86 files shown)


src/client/components/common/Button/index.tsx
```diff
@@ -14,7 +14,7 @@ const StyledButton = styled.button<{ outline?: boolean }>`
   flex-direction: row;
   align-items: center;
   justify-content: center;
-  padding: 0 1rem;
+  padding: 0 1.6rem;
```
turtlemoji, line 17: I couldn't find a place where the buttons had 16px of paddng in the figma (they are custom widths mostly), but mostly important this component affects all the buttons in the app, and they should not have this amount of padding.

src/client/components/common/Card/CardElement/index.tsx
```diff
@@ -1,9 +1,9 @@
 import { FC, ReactNode } from 'react';
-import styled from 'styled-components';
+import styled from 'styled-components/macro';
 
 import { Icon, ArrowDownIcon, IconProps } from '@components/common';
 
-const Container = styled.div<{ width?: string; align?: string; grow?: string; fontWeight?: number }>`
+const Container = styled.article<{ width?: string; align?: string; grow?: string; fontWeight?: number }>`
```
turtlemoji, line 6: The CardElement is a wrapper for all the elements inside card, including every element here:
image

I think we should leave it as a div

src/client/components/common/Card/CardElement/index.tsx
```diff
@@ -34,16 +34,18 @@ const SortIcon = styled(({ activeSort, sortType, ...props }: SortIconProps) => <
...
   font-size: 1.6rem;
+  font-weight: 400;
+  margin: 0 0 0.8rem 0;
+  padding: 0;
   color: ${({ theme }) => theme.colors.texts};
   cursor: ${({ onClick }) => (onClick ? 'pointer' : 'default')};
 `;
 
 const Content = styled.div`
-  margin-top: 0.8rem;
```
turtlemoji, line 46: It should have it normally. Or maybe you find another case ?

src/client/components/common/Card/CardHeader/index.tsx
```diff
@@ -1,9 +1,10 @@
 import { FC } from 'react';
-import styled from 'styled-components';
+import styled from 'styled-components/macro';
 
 const Container = styled.div`
   width: 100%;
   padding: 0 ${({ theme }) => theme.card.padding};
+  margin-bottom: 1.2rem;
```
turtlemoji, line 7: Same here, We can have a cardHeader without content below, and this can create a weird space below it

src/client/components/common/Button/index.tsx
turtlemoji: I think the current one should be layoutPadding / 2 (8px)

src/client/components/common/Card/CardElement/index.tsx
pentcle: an article is a self-contained element - so a card is a good use of article: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

src/client/components/common/Card/index.tsx
```diff
@@ -83,7 +83,7 @@ export interface CardProps extends StyledSystemProps {
   cardSize?: CardSizeType;
 }
 
-const StyledDiv = styled.div<CardProps>`
+const StyledDiv = styled.article<CardProps>`
```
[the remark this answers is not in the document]
pentcle: article is generally what i'd use as a self-contained wrapper for a small amount of repeatable content
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

src/client/components/common/Card/CardElement/index.tsx
turtlemoji: On a card yes, but in this cases the CardElement is not a card, it is a wrapper not related to the content itself

xgambitox:
@pentcle looks good overall! Would just take a look at @turtlemoji comments to make sure all ok

src/client/components/common/Button/index.tsx
turtlemoji: I would fix the connectwalletbutton width to a min width, and left default horizontal padding layoutPadding / 2

turtlemoji: Also remember mobile, since we don't have the designs yet, leaving big paddings will just create more overhead for us to add media queries

pentcle:
went off on an exploration of themes - just pushing minimum for now - the dark theme (buttons, logo) don't match the wireframes - so back to working on that now. have also noticed the wireframes don't match the design library in the doc - so checking on which is correct

src/client/components/common/Button/index.tsx
pentcle: cool - have reverted for now

src/client/components/common/Card/CardRedirection/index.tsx
```diff
@@ -1,13 +1,13 @@
...
-import styled from 'styled-components';
+import styled from 'styled-components/macro';
 
 // NOTE: App errors out when trying to import directly from 'common'
 import { ChevronRightIcon, Icon } from '@components/common/Icon';
 
 const Arrow = styled(Icon)`
   height: 1.6rem;
-  fill: currentColor;
+  fill: ${({ theme }) => theme.colors.primary};
```
turtlemoji, line 10: this overrides inheritance for the component itself

src/client/components/common/OptionList.tsx
```diff
@@ -45,7 +45,7 @@ const StyledSpinnerLoading = styled(SpinnerLoading)`
 
 const ArrowIcon = styled(Icon)<{ open?: boolean }>`
   margin-left: 0.8rem;
-  height: 1rem;
+  height: 0.8rem;
```
turtlemoji, line 48: image
I'm seeing this on figma

src/client/components/app/Navbar/index.tsx
```diff
@@ -15,21 +15,21 @@ const StyledOptionList = styled(OptionList)`
 const StyledNavbarActions = styled.div`
   display: grid;
   grid-auto-flow: column;
-  grid-gap: 0.8rem;
+  grid-gap: 1.6rem;
```
turtlemoji, line 18: we should check all buttons and mobile in another pr later, just to have it in mind

src/client/components/common/Card/CardRedirection/index.tsx
pentcle: i checked it in the other themes - and it's meant to be blue in the mono theme

src/client/components/common/OptionList.tsx
pentcle: haha

src/client/components/common/Card/CardElement/index.tsx
```diff
@@ -51,6 +54,7 @@ const Content = styled.div`
   text-overflow: ellipsis;
   overflow: hidden;
   max-width: 100%;
+  margin-top: 0.8rem;
```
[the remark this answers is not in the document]
pentcle: ok - hmm - it means we cannot have the card designed the way we want to - maybe some of these components may want to be a bit more specific - but we can talk about that in next round of work - i shall comply : )

PR #565, opened 2022-03-29
fix: console errors on open modal

### LINKED ISSUE #508: fix: lint errors on modal
**Describe the bug**
During investigation of double approve came across new lint errors revolving around loading the modal

**To Reproduce**
1.  Open the modal and some errors will pop out.
Ex:
image

huggingbot:
Added React Fragments to several JSX Elements as a TransitionGroup expects its immediate children to be of type Transition. It will pass in `onExit` props to all of those child nodes. The errors appear as those properties are not valid to the child nodes. If we wrap the components in a Transition or CSSTransition or a React Fragment, those warnings will go away.

Fixes #508

### DIFF
--- a/src/client/components/app/Transactions/Transaction.tsx
+++ b/src/client/components/app/Transactions/Transaction.tsx
@@ -109,7 +109,7 @@ export const Transaction: FC<TransactionProps> = (props) => {
 
   if (transactionCompleted) {
     return (
-      <StyledTransaction onClose={onClose} header={transactionLabel} {...props}>
+      <StyledTransaction onClose={onClose} header={transactionLabel}>
         <TxStatus transactionCompletedLabel={transactionCompletedLabel} exit={onTransactionCompletedDismissed} />
       </StyledTransaction>
     );
@@ -130,7 +130,7 @@ export const Transaction: FC<TransactionProps> = (props) => {
   }`;
 
   return (
-    <StyledTransaction onClose={onClose} header={transactionLabel} {...props}>
+    <StyledTransaction onClose={onClose} header={transactionLabel}>
       <TxTokenInput
         headerText={sourceHeader}
         inputText={sourceInputText}
--- a/src/client/components/app/Transactions/components/TxTokenInput.tsx
+++ b/src/client/components/app/Transactions/components/TxTokenInput.tsx
@@ -265,7 +265,7 @@ export const TxTokenInput: FC<TxTokenInputProps> = ({
 
   return (
     <StyledTxTokenInput {...props}>
-      {headerText && <Header>{headerText}</Header>}
+      <>{headerText && <Header>{headerText}</Header>}</>
       {openedSearch && (
         <CSSTransition in={openedSearch} appear={true} timeout={scaleTransitionTime} classNames="scale">
           <StyledSearchList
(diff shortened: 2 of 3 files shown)


src/client/components/app/VaultDetail/VaultDetailPanels.tsx
```diff
@@ -408,8 +408,8 @@ export const VaultDetailPanels = ({
               <StyledCardHeader header={t('vaultdetails:overview-panel.strategies')} />
 
               <OverviewStrategies>
-                {strategies.map((strategy) => (
-                  <OverviewInfo variant="surface" cardSize="micro">
+                {strategies.map((strategy, index) => (
+                  <OverviewInfo variant="surface" cardSize="micro" key={index}>
```
karelianpie, line 412: We should avoid using index as key in this case because we do have stable IDs in the form of the strategy address

> When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort

https://reactjs.org/docs/lists-and-keys.html#keys

```suggestion
                  <OverviewInfo variant="surface" cardSize="micro" key={strategy.address}>
```

src/client/components/app/Transactions/Transaction.tsx
```diff
@@ -109,7 +110,7 @@ export const Transaction: FC<TransactionProps> = (props) => {
 
   if (transactionCompleted) {
     return (
-      <StyledTransaction onClose={onClose} header={transactionLabel} {...props}>
+      <StyledTransaction onClose={onClose} header={transactionLabel} {...restProps}>
```
karelianpie, line 113: Wondering what props are in the `restProps`?

huggingbot: As of now `restProps` is an empty object as `StyledTransaction` only has 2 props which are `header` and `onClose`. `restProps` can be removed to prevent future developers from mistakenly pass in incorrect props, and would require them to explicitly pass in the prop to `StyledTransaction`.

karelianpie: Yeah, that's what I thought as well... that there wouldn't be anything in there. Happy to remove them

src/client/components/app/VaultDetail/VaultDetailPanels.tsx
karelianpie: Addressed with https://github.com/yearn/yearn-finance-v3/pull/565/commits/0f6d97fd2f804a2b5b5450e68e7460eff088ec16

src/client/components/app/Transactions/components/TxTokenInput.tsx
```diff
@@ -278,40 +278,42 @@ export const TxTokenInput: FC<TxTokenInputProps> = ({
...
-            )}
-            {yieldPercent && (
-              <StyledText>
-                {t('components.transaction.token-input.yield')} <ContrastText>{yieldPercent}</ContrastText>
-              </StyledText>
-            )}
-          </TokenExtras>
-        </TokenData>
-      </TokenInfo>
+      <>
```
karelianpie, line 281: Are we sure these are needed now? Usually fragment is used when we want to group a list of children without adding extra nodes to the DOM.

karelianpie: That makes sense, just wondering if there's a better way. I guess we can go with that for now.

karelianpie (approved):
LGTM! Still wondering if there's a better way than using Fragments though.

Thank you for the contribution @huggingbot 🙇

xgambitox (approved):

PR #569, opened 2022-03-30
test: `BigNumber` formatters

karelianpie:
Since I've added the same formatters in the SDK (https://github.com/yearn/yearn-sdk/pull/267), thought of reusing the tests in here with the added benefit of getting testing the formatters started.

Tests the following utility functions that work with `BigNumber`

* `toBN()`
* `toWei()`
* `toUnit()`

### DIFF
--- a/src/utils/format.spec.ts
+++ b/src/utils/format.spec.ts
@@ -1,0 +1,62 @@
+import BigNumber from 'bignumber.js';
+
+import { toBN, toUnit, toWei } from './format';
+
+describe('Fromatters', () => {
+  describe('toBN', () => {
+    it('should convert `string` to `BigNumber`', () => {
+      const actual = toBN('123');
+
+      expect(actual).toEqual(new BigNumber(123));
+      expect(actual instanceof BigNumber).toBeTruthy();
+    });
+
+    it('should convert `number` to `BigNumber`', () => {
+      const actual = toBN(123);
+
+      expect(actual).toEqual(new BigNumber(123));
+      expect(actual instanceof BigNumber).toBeTruthy();
+    });
+
+    it.each`
+      falsyArgument
+      ${false}
+      ${0}
+      ${''}
+      ${null}
+      ${undefined}
+      ${NaN}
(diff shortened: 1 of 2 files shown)


src/utils/format.spec.ts
```diff
@@ -0,0 +1,62 @@
+import BigNumber from 'bignumber.js';
+
+import { toBN, toUnit, toWei } from './format';
+
+describe('Fromatters', () => {
```
turtlemoji, line 5: "formatters"

src/utils/format.ts
```diff
@@ -24,8 +24,7 @@ const FORMAT = {
 /* -------------------------------------------------------------------------- */
 
 export const toBN = (amount?: Amount | number): BigNumber => {
-  if (!amount || amount === '') amount = '0';
-  return new BigNumber(amount);
+  return new BigNumber(amount || 0);
```
turtlemoji, line 27: are we sure we want to return a bn of 0 instead of throwing an error here? @xgambitox  @karelianpie

src/utils/format.spec.ts
karelianpie: Fixed with https://github.com/yearn/yearn-finance-v3/pull/569/commits/72cb3529229e5813175c4919f1d25ac37f89649f

src/utils/format.ts
karelianpie: What the actual `BigNumber` from 'bignumber.js' does for example in case its given a `null` is
```typescript
new BigNumber(null) // BigNumber { s: null, e: null, c: null }
```
I think big number of 0 is sensible for us, but I don't have a strong opinion here. The change above was just a refactor, the logic remained as it was, just saving a line since `amount || 0` is the same as what the other line was doing. I'm happy to refactor to throw if you Turtle or Gambito have a strong opinion either way.

xgambitox: Yes its intentional. We save a lot of lines of code in many places where amount might be undefined, and then we dont need to place safeguards to continue with the math operation.

xgambitox (approved):

turtlemoji (approved):

PR #579, opened 2022-04-05
build(browserslist): update config

sambacha:
this configuration for browserslist explicitly drops IE support and unsafe browser versions. This should reduce build size/polyfill usage in production compiled output

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "yearn-finance-v3",
-  "version": "1.0.32",
+  "version": "1.0.33",
   "private": true,
   "scripts": {
     "start": "craco start",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -98,14 +93,22 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
+      ">0.3%",
+      "not IE > 0",
+      "not op_mini all",
       "not dead",
-      "not op_mini all"
+      "last 2 Edge major versions",
```
turtlemoji, line 100: @sambacha Should we really drop de polyfills for almost everything? We will destroy the styling in not updated browsers. How much speed we can get from removing this pollys?

turtlemoji: Maybe I'm mistaken in how the config works (include / exclude) :)

turtlemoji: With this config we will have polys for:

>0.3% &&
last 2 major versions (even if they are not in the > 0.3%) &&
Chrome 99 but exclude ios safari 12.5?

If that's the case then I think it's ok and I didn't interpret it ok the first time

turtlemoji (approved):

PR #584, opened 2022-04-09
fix: eth deposit modal

huggingbot:
Fixes #571 

Added `createPlaceholderToken` in `misc.ts` to create placeholder tokens.

If the user has no ETH in their wallet, the placeholder token will be used provided the selected sell token is the native currency.

PR #586, opened 2022-04-12
fix: always show eth

FiboApe:
Need to be merged after release of https://github.com/yearn/yearn-sdk/pull/278


xgambitox:
Since we should expect same issue to happen in other networks, we should refactor in other PR all the app where we use ETH address directly, and use from now on `0x00...` as native token address.

turtlemoji (approved):

turtlemoji:
hold until SDK release

FoxTheSin:
functionality appears good but UI stuff is old

FiboApe:
Closing this in favor of #584

PR #599, opened 2022-04-16
build: upgrade dependencies

turtlemoji:
We still need to upgrade react-scripts, seems that it can't be done easily @karelianpie, need to check further but the build needs it.
@FoxTheSin can you test views and some basic interaction just to be sure that this is ok?

The goal of this updates to the dependencies, is for security and updating subdependencies in yarn.lock on another pr.


FoxTheSin:
desktop themes looks good
fantom deposit & withdraw look good (fbeets)

FoxTheSin (approved):

karelianpie (approved):
LGTM

PR #601, opened 2022-04-18
build: update react-scripts

turtlemoji:


turtlemoji:
No errors on build, but we are not rendering anything in the page and we have this error on console:
image

turtlemoji:
Hold this until craco is compatible with CRA 5 or think if it's better to drop craco.

PR #607, opened 2022-04-21
feat: add new design for errors on deposit modal [WEB-1466] 

FiboApe:
Screenshots:
Screenshot 2022-04-21 at 20 07 07
Screenshot 2022-04-22 at 13 42 36
Screenshot 2022-04-22 at 13 42 58


src/client/components/app/Transactions/components/TxError.tsx
```diff
@@ -1,62 +1,118 @@
...
-      <StyledIcon Component={WarningIcon} />
+      <StyledTxErrorTitle>
+        <StyledIcon Component={WarningIcon} errorType={errorType} />
+
+        <StyledTextTitle>{errorTitle || t('errors.unknown')}</StyledTextTitle>
 
-      <StyledText>{errorText || t('errors.unknown')}</StyledText>
+        {errorDescription && (
+          <StyledIcon
+            onClick={() => setIsExpanded(!isExpanded)}
```
karelianpie, line 105: This might lead to unexpected results because this is asynchronous, we should use the previous state rather setting this directly

```suggestion
            onClick={() => setIsExpanded((prevIsExpanded) => !prevIsExpanded)}
```

src/client/themes/cyberpunk/index.ts
```diff
@@ -69,8 +69,14 @@ const cyberpunkTheme: DefaultTheme = {
       primary: '#00A3FF',
       onPrimary: '#FFFFFF',
       loading: '#FFA800',
-      error: '#FF005E',
-      warning: '#FF7500',
+      error: {
+        background: '#FF005E',
```
karelianpie, line 73: Is this the background-color? I think if so, we should try and use `backgroundColor` instead, unsure if that's a big refactor

src/utils/validations.ts
```diff
@@ -111,11 +111,11 @@ export function validateVaultWithdraw(props: ValidateVaultWithdrawProps): Valida
   if (yvTokenAmount.isZero()) return {};
 
   if (amountInWei.lt(0)) {
-    return { error: 'INVALID AMOUNT' };
+    return { error: 'Invalid amount' };
```
karelianpie, line 114: I feel these errors could be a bit more meaningful if we were to use the logic to get here. Example:
```suggestion
    return { error: `Invalid amount: ${amountInWei}` };
```

src/client/components/app/Transactions/components/TxContainer.tsx
```diff
@@ -62,7 +62,6 @@ const StyledTxContainer = styled.div`
   flex-direction: column;
   padding: ${({ theme }) => theme.card.padding};
   width: 100%;
-  height: 53rem;
```
turtlemoji, line 65: This will cause rerendering and wrong style in most of the txcontainers, we should leave it or change the height to another value to reflect the new changes and don't cause an overflow

src/client/components/app/Transactions/components/TxError.tsx
```diff
@@ -1,62 +1,118 @@
...
-import { Icon, Text, WarningIcon } from '@components/common';
+import { ChevronDownIcon, ChevronUpIcon, Icon, Text, WarningIcon } from '@components/common';
 import { useAppTranslation } from '@hooks';
 
-const StyledIcon = styled(Icon)`
-  width: 4rem;
-  fill: ${({ theme }) => theme.colors.txModalColors.textContrast};
+const StyledIcon = styled(Icon)<{ errorType?: ErrorType }>`
+  width: 2rem;
+  fill: ${({ theme }) => theme.colors.txModalColors.error.color};
```
turtlemoji, line 9: we just do fill: currentColor, and color: inherit, so you will not need to specify it again here and only on the parent

src/client/themes/dark/index.ts
```diff
@@ -107,8 +107,14 @@ const darkTheme: DefaultTheme = {
       primary: '#00A3FF',
       onPrimary: '#FFFFFF',
       loading: '#FFA800',
-      error: '#FF005E',
-      warning: '#FFA800',
+      error: {
```
turtlemoji, line 110: we have this too, as global vars in default file. Don't know if it should change between themes or not, I will leave it at your criteria

turtlemoji: image

src/utils/validations.ts
```diff
@@ -111,11 +111,11 @@ export function validateVaultWithdraw(props: ValidateVaultWithdrawProps): Valida
...
   }
 
   if (amountInWei.gt(userYvTokenBalance)) {
-    return { error: 'INSUFFICIENT FUNDS' };
+    return { error: 'Insufficient funds' };
```
turtlemoji, line 118: we should create another PR in the future and have error codes or something like "INSUFFICIENT_FUNDS" so we can just translate them and have them all in a .json file

turtlemoji:
@FiboApe can you add another error component in the test modal with description so we can try the expand more easily?
image

src/utils/validations.ts
FiboApe: in this case amountInWei would be just what the user set on the input, I don't know if it makes much sense to tell the user that

src/client/components/app/Transactions/components/TxContainer.tsx
FiboApe: Height should be defined by the content and not the txcontainer (maybe we can set a max one), which of the other contents would make this break?

src/client/themes/dark/index.ts
FiboApe: But those are for alerts, this is the modal, should we use the same for both?

src/client/components/app/Transactions/components/TxError.tsx
FiboApe: fixed on 43649ef98fa2d6e9d704f21f6f7892a3ba13f7f2

src/client/themes/cyberpunk/index.ts
FiboApe: fixed on 43649ef98fa2d6e9d704f21f6f7892a3ba13f7f2

src/client/components/app/Transactions/components/TxError.tsx
FiboApe: solved on 5c0f6e767314e1ad4f1dc71847736db858f9970f

src/client/components/app/Transactions/components/TxContainer.tsx
turtlemoji: In the designs the height was fixed, and it was not defined by the content in all cases. More like having a max and min height. If not, it will overflow in the vault details, or cause rerenders

src/client/themes/dark/index.ts
turtlemoji: Idk, depending if the colors are ok, in reality that colors are just global variables, not only for alerts itself. It depends if the error colors should change betweeen themes I suppose

turtlemoji:
@FiboApe should we create another PR for the transaction messages? (Success, loading, etc)
image

turtlemoji:
@FiboApe can you bump package minor version?

src/client/components/app/Transactions/components/TxContainer.tsx
turtlemoji: image
In this case, for exxample, we should be flexing the content to the bottom to occupy all the remaining space

src/utils/validations.ts
karelianpie: Good point 😅 I forgot these were coming from there. All good, I had a look at the others and the user should be able to deduce why the error is happening. Resolving!

src/client/components/app/Transactions/TestTx.tsx
```diff
@@ -12,6 +12,13 @@ export interface TestTxProps {
 
 const StyledTestTx = styled(TxContainer)``;
 
+const LONG_DESCRIPTION =
```
karelianpie, line 15: What's this?

karelianpie (approved):
Is it me or this would be nicer with a little bit more padding around the error message? Text seems a bit too close to the edge

feat__add_new_design_for_errors_on_deposit_modal__WEB-1466__by_FiboApe_·_Pull_Request__607_·_yearn_yearn-finance-v3

Otherwise, LGTM. Thanks for adding the screenshots! 👍

turtlemoji:
> Is it just me or this would be nicer with a little bit more padding around the error message?
> 
> Text seems a bit too close to the edge:
> 
> feat__add_new_design_for_errors_on_deposit_modal__WEB-1466__by_FiboApe_·_Pull_Request__607_·_yearn_yearn-finance-v3
> 
> Otherwise, LGTM. Thanks for adding the screenshots! 👍

image
 
 This is what design delivered so I will follow it for now just to don't stall waiting for their response, lgtm

karelianpie:
feat__add_new_design_for_errors_on_deposit_modal__WEB-1466__by_FiboApe_·_Pull_Request__607_·_yearn_yearn-finance-v3

@turtlemoji when I overlay it appears the design has a bit more padding at the top at least, looks like another 8px perhaps

karelianpie (changes requested):
Sorry, I think we should be closer to the designs, the error message needs a bit more breathing room

turtlemoji:
Yes, if is different lets change it :) @FiboApe

turtlemoji:
ok, matched design with the figma, (but removed .8 of margin-top because it looked really bad in the implementation, figma used 1.6rem for margin-top and I left only .8rem)
Moved click to all the title so it works on mobile (too difficult to click the dropdown icon)
Also changed the icon to a new one used in the figma (yeah, it is different)

67c055b5b68004d474b2dac59d2b6c8f4668c9e0

turtlemoji:
image

turtlemoji:
Also @FiboApe , I think we will have to change the colors for some themes:
image

karelianpie:
> ok, matched design with the figma, (but removed .8 of margin-top because it looked really bad in the implementation, figma used 1.6rem for margin-top and I left only .8rem)

Fair enough, thanks! I think it looks good now 🎉 

> Moved click to all the title so it works on mobile (too difficult to click the dropdown icon)

Good idea! 👍

src/client/components/app/Transactions/components/TxError.tsx
```diff
@@ -1,62 +1,151 @@
...
+  fill: currentColor;
+  flex-shrink: 0;
+  transform: rotate(180deg);
+  transition: transform 200ms ease-in-out;
+
+  ${({ isExpanded }) =>
+    isExpanded &&
+    `
+    transform: rotate(0);
+  `}
```
karelianpie, line 18: nit: Is this how lint/prettier prefers it? Looks quite hard to read

karelianpie (approved):

src/client/components/app/Transactions/components/TxError.tsx
turtlemoji: Yeah. 
I would love this
image

But prettier gets really angry at that.
Maybe @FiboApe can create/edit a rule for this? He is really good with lint rules

turtlemoji (changes requested):
fix cyberpunk colors of txError

src/client/components/app/Transactions/components/TxError.tsx
karelianpie: 100%, that would be the way to format that in a much more readable way

src/client/themes/dark/index.ts
FoxTheSin: For ease sake, keep it the same. After release (because I already freaking love this update to the errors) we can update UI for cleaner/easier reading. Function first IMO.

src/client/components/app/Transactions/TestTx.tsx
```diff
@@ -13,11 +13,7 @@ export interface TestTxProps {
...
...andit, neque ligula viverra eros, eu volutpat nulla massa a elit. Sed a sem vitae purus hendrerit consequat sed id urna. Vivamus convallis quam mi, vel rutrum augue suscipit ut. Nunc vel ornare metus, a commodo justo. Morbi vel auctor turpis. Vestibulum id lacinia nulla. In tincidunt sodales iaculis. Sed elementum hendrerit quam, iaculis lobortis lacus euismod non. Quisque ligula arcu, varius sit amet bibendum pulvinar, mattis vitae justo. Mauris at nisi eu orci tristique tempor id a nibh. Etiam sollicitudin sollicitudin dictum. Cras nec urna et nibh aliquam condimentum eget in lacus. Fusce ornare, ligula sed aliquam scelerisque, arcu dui placerat lacus, nec mattis ante orci at enim. Fusce lobortis risus neque. Donec mollis sit amet nulla in egestas. Pellentesque a ipsum sed sapien posuere tincidunt. Integer eget libero eros. Fusce vitae luctus quam, ac blandit leo. Cras ullamcorper viverra magna vel malesuada. In ac mollis leo. Cras aliquam erat imperdiet felis placerat malesuada. Aliquam a scelerisque nunc. Cras et bibendum magna. Fusce eros ex, tincidunt a diam et, dignissim mollis nulla. In sagittis sed diam luctus hendrerit. Nunc posuere odio sit amet arcu pharetra pulvinar.';
```
FiboApe, line 16: this was actually formatted by prettier

turtlemoji: Yeah, idk why it break the build in one stage

karelianpie (approved):

src/client/components/app/Transactions/components/TxError.tsx
```diff
@@ -1,62 +1,151 @@
...
+    margin: .8rem 0;
+  `}
+
+  ::-webkit-scrollbar {
+    width: 0.2rem;
+  }
+
+  /* Track */
+  ::-webkit-scrollbar-track {
+    background: ${({ theme }) => theme.colors.txModalColors.error.color}80;
```
turtlemoji, line 91: typo

src/client/components/app/Transactions/components/TxError.tsx
```diff
@@ -1,62 +1,151 @@
...
+    background: ${({ theme }) => theme.colors.txModalColors.error.color};
+  }
 
   ${({ errorType, theme }) =>
     errorType === 'warning' &&
     `
-    background-color: ${theme.colors.txModalColors.warning};
+      /* Track */
+      ::-webkit-scrollbar-track {
+        background: ${theme.colors.txModalColors.warning.color}80;
```
turtlemoji, line 104: typo?

src/client/components/app/Transactions/components/TxError.tsx
```diff
@@ -1,62 +1,151 @@
...
   errorType?: ErrorType;
 }
 
-export const TxError: FC<TxErrorProps> = ({ errorText, errorType, children, ...props }) => {
+export const TxError: FC<TxErrorProps> = ({ errorTitle, errorDescription, errorType, children, ...props }) => {
   const { t } = useAppTranslation('common');
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  useEffect(() => {
+    //clean state once error message changes
```
turtlemoji, line 126:
```suggestion
    // NOTE Clean state once error message changes
```

src/client/components/app/Transactions/components/TxError.tsx
FiboApe: nope, it adds 80% for opacity

src/client/components/app/Transactions/components/TxError.tsx
FiboApe: nope, it adds 80% for opacity

FiboApe:
@turtlemoji fixed height issue and allowed it to have variable height here: e6ee4b6b92d8cf2ca25b28ae29ab1425fd69798d
Screenshot 2022-04-28 at 14 39 01
Screenshot 2022-04-28 at 14 39 13
Screenshot 2022-04-28 at 14 39 21
Screenshot 2022-04-28 at 14 39 27

karelianpie (approved):

src/client/components/common/Tabs.tsx
```diff
@@ -103,6 +103,7 @@ export interface TabPanelProps extends StyledSystemProps {
 }
 
 const StyledTabPanel = styled.div<{ hidden?: boolean }>`
+  ${({ hidden }) => (hidden ? 'display: none;' : 'display: flex;flex: 1;')}
```
turtlemoji, line 106: @FiboApe missing space after `;`. I like to not use the ternary tho and just do more lines in this case

turtlemoji (approved):

src/client/components/app/Transactions/components/TxError.tsx
turtlemoji: the thing is, what appens if the variable is an RGB, it will break, this only works with hexas

xgambitox: should we rather set the opacity separately ?

src/client/components/app/Transactions/components/TxError.tsx
xgambitox: should we rather set the opacity separately ?

xgambitox: should we rather set the opacity separately ?

xgambitox (approved):

karelianpie (approved):

src/client/components/app/Transactions/components/TxError.tsx
FiboApe: I can add a new constant to the theme just to have it as opacity, but there is no way to set opacity of the background separately

turtlemoji: you can always create a function to check if the color is an RGB, RGBA, or an hexa, and then apply the opacity depending that lol. 
I would just not use opacity in this case and not have problems in the future, and just try with another var. like mixing background color and surface color.

PR #617, opened 2022-04-28
fix: order labs by highest apy

### LINKED ISSUE #615: Lab highest apy % are not ordered correctly
Screenshot 2022-04-27 at 17 12 07

yvBoost is in the middle

edwinzhng:
## Description

- Sorts the labs vaults by highest APY when displaying

## Related Issue

Fixes #615

## How Has This Been Tested?

Tested locally by checking the labs page

Screen Shot 2022-04-27 at 11 20 34 PM

### DIFF
--- a/src/client/routes/Labs/index.tsx
+++ b/src/client/routes/Labs/index.tsx
@@ -306,14 +306,14 @@ export const Labs = () => {
           {currentNetworkSettings.labsEnabled && (
             <StyledRecommendationsCard
               header={t('components.recommendations.header')}
-              items={recommendations
-                .sort((a, b) => Number(b.apyData) - Number(a.apyData))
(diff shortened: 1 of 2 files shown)


edwinzhng:
This will conflict with #561 so I can fix any issues after it gets merged

xgambitox:
What do you think if we handle the sorting in the selector as we do in vaults [here](https://github.com/yearn/yearn-finance-v3/blob/develop/src/core/store/modules/vaults/vaults.selectors.ts#L153) ?

That way if we use the labs recommendations in other place in the future, it will be sorted too

edwinzhng:
thanks! good catch, I can move the sort

karelianpie (dismissed):
LGTM 👍

karelianpie (approved):

PR #622, opened 2022-04-29
feat: support for web-lib

Majorfi:
## Description
Add a basic support for the weblib. The lib is integrated not to disturb how the app is currently working, but should provide access to all the components currently available on the weblib.
You can find the lib source code [here](https://github.com/yearn/web-lib), and the documentation [here](https://yearn.systems/doc/components).

Please see comments directly on the code for explainations.

Note about css: the weblib uses tailwindcss v3. We had some talk about bringing it to the project and it was decided not to right now. To access the styling of the lib, only the css relevant to the weblib is extracted from it, minified and cleaned, then imported as static file. This will be integrated as a lib-level css file in one of the next version of the lib.
Note about em/rem: the yearn.finance website uses a hack to scale font that makes the `rem` unit not really easy to use. The exported CSS uses `em` instead of the classic `rem`.

## Related Issue
*N/A*

## Motivation and Context
Trying to merge effort between the multiple frontends and only do the work once.

## How Has This Been Tested?
No breaking change, only some theming. Tested to fix some issues with the Amsterdam Banner component and the theming system

*NA*


public/index.html
```diff
@@ -40,7 +40,7 @@
 
     <script async defer data-domain="beta.yearn.finance" src="https://analytics.yearn.finance/js/plausible.js"></script>
   </head>
-  <body>
+  <body class='transition-colors duration-150 bg-background' data-theme='system-prefs'>
```
Majorfi, line 43: The class could be removed, but: 
- `transition-colors duration-150` makes the transition between theme a bit smoother
-  `bg-background` overwrite the default background color (even on overscroll) by the one currently selected via the theming system

The `data-theme` argument is used to indicate the current theme and be able to play with it in the css/js. Currently, 3 themes are supported:
- `system-prefs` -> `dark` or `light` depending on the user's preference
- `dark` -> well, dark theme
- `light` -> yes, light theme

The custom themes are **NOT** supported by the lib yet, we should add them

src/client/assets/images/index.ts
```diff
@@ -2,6 +2,5 @@ import AlchemyCertified from './alchemy-certified.svg';
 import YearnLogoFull from './yearn-logo-full.svg';
 import YearnLogo from './yearn-logo.svg';
 import MetamaskLogo from './metamask-logo.png';
-import AmsterdamLogo from './amsterdam-banner.jpg';
 
-export { AlchemyCertified, YearnLogoFull, YearnLogo, MetamaskLogo, AmsterdamLogo };
```
Majorfi, line 7: For testing purpose, I used the AmsterdamBanner as example of integration. Because of that, we no longer need that

src/client/routes/Portfolio/index.tsx
```diff
@@ -20,27 +21,20 @@ import {
...
 
-const AmsterdamButton = styled(Button)`
-  background: #0657f9;
-  color: #fff;
-`;
-
-const StyledSliderCard = styled(SliderCard)`
+const AmsterdamCard = styled.div`
   grid-column: 1 / 3;
 `;
```
Majorfi, line 39: Creating a styled.div to fix the whole width

src/client/routes/Portfolio/index.tsx
```diff
@@ -176,21 +170,9 @@ export const Portfolio = () => {
...
-                <AmsterdamButton>Follow @iearnfinance</AmsterdamButton>
-              </a>
-            </p>
-          </Text>
-        }
-        background={<img src={AmsterdamLogo} alt="Amsterdam logo"></img>}
-      />
+      <AmsterdamCard>
+        <BannerEventAmsterdam />
+      </AmsterdamCard>
```
Majorfi, line 175: Using the Component \o

src/core/store/modules/theme/theme.reducer.ts
```diff
@@ -15,6 +15,8 @@ const themeReducer = createReducer(themeInitialState, (builder) => {
   builder.addCase(changeTheme.fulfilled, (state, { payload: { theme } }) => {
     // TODO Check that the theme exists on AVAILABLE_THEMES or AVAILABLE_CUSTOM_THEMES
     state.current = theme;
+    //Adding the current theme to the dataset of body, in order to use it with css queries
+    document.body.dataset.theme = theme;
```
Majorfi, line 19: Put the currently selected theme in `data-theme` of the body. Used to track the current theme.

package.json
```diff
@@ -42,6 +42,7 @@
     "@types/redux-logger": "3.0.9",
     "@types/styled-components": "5.1.12",
     "@types/styled-system": "5.1.13",
+    "@yearn/web-lib": "^0.2.24",
```
karelianpie, line 45: We probably should lock this version even though it's an internal package

karelianpie (dismissed):
LGTM, probably some of the new config in `.github/workflows/nodejs.yml` is not letting it build, I unfortunately don't have access to AWS to see why it's failing

src/client/index.tsx
```diff
@@ -28,8 +28,8 @@ const GlobalStyle = createGlobalStyle`
   }
 
   body {
-    background-color: ${(props) => props.theme.colors.background};
-    color: ${(props) => props.theme.colors.texts};
+    background-color: rgb(var(--color-background));
+    color: rgb(var(--color-typo-secondary));
```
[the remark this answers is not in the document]
turtlemoji: this will fuck up the ui colors, since you will change this ones, but not all the other components and etc ones

src/client/routes/Portfolio/index.tsx
turtlemoji: Removed the banner tho

src/client/routes/Portfolio/index.tsx
turtlemoji: removed in latest release

package.json
turtlemoji: yes, lock

package.json
```diff
@@ -61,7 +62,7 @@
     "lodash": "4.17.21",
     "prettier": "2.6.2",
     "react": "17.0.2",
-    "react-countup": "5.2.0",
+    "react-countup": "^6.2.0",
```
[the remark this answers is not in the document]
Majorfi: Fixed with [3809437](https://github.com/yearn/yearn-finance-v3/pull/622/commits/38094379ce6b428518442ff9fb1db52c5ed577c7)

package.json
Majorfi: Fixed with [5d07fe3](https://github.com/yearn/yearn-finance-v3/pull/622/commits/5d07fe342ecf54d9d45aefc2c116073f462ce185)

src/client/index.tsx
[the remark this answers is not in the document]
Majorfi: Removed with [4046d88](https://github.com/yearn/yearn-finance-v3/pull/622/commits/4046d887a1bb3f83a70c9e945d03500d5faf722b)

This should be correctly handled by the code anyway

package.json
```diff
@@ -61,7 +62,7 @@
     "lodash": "4.17.21",
     "prettier": "2.6.2",
     "react": "17.0.2",
-    "react-countup": "5.2.0",
+    "react-countup": "6.2.0",
```
turtlemoji, line 65: why do we need this one tho? Can't you add it to the lib?

turtlemoji, line 65: why do we need this one tho? Can't you add it to the lib?

public/index.html
xgambitox: Can we try to add these styles directly in the global style [here](https://github.com/yearn/yearn-finance-v3/blob/develop/src/client/index.tsx#L14) ?

a reference from styled components docs [here](https://styled-components.com/docs/api#createglobalstyle)

Just to try to keep globally applied css in one place.

src/core/store/modules/theme/theme.reducer.ts
xgambitox: should we better set it dynamically in the global styles using the theme selector ? 

Is not a good practice to modify other values out of the scope of the reducer (see docs [here](https://redux.js.org/style-guide/#reducers-must-not-have-side-effects))

karelianpie:
@Majorfi I've asked in YFI Infrastructure why the AMS Amplify is failing. Will let you know

src/core/store/modules/theme/theme.reducer.ts
Majorfi: I found it like the easiest way to be sure the data is always updated on theme change. I would be happy to change that, but do you have an _exact_ idea where we should put it ? 😅

public/index.html
Majorfi: Fixed with [5b1ee4e](https://github.com/yearn/yearn-finance-v3/pull/622/commits/5b1ee4e84b83871a45aff82c84a66a41a98b5aa1) ser!

src/core/store/modules/theme/theme.reducer.ts
xgambitox: Yeah, we can place it in the action for now, as its a better place for it. 

Maybe the best place would be moving these [GlobalStyle](https://github.com/yearn/yearn-finance-v3/blob/develop/src/client/index.tsx#L82) component into the [Themable](https://github.com/yearn/yearn-finance-v3/blob/develop/src/client/containers/Themable/index.tsx) container, so that we can access the current selected theme state, and pass it down as a prop to be used in the html body. We can spend more time later on that refactor though.

xgambitox:
> @Majorfi I've asked in YFI Infrastructure why the AMS Amplify is failing. Will let you know

Lets try running `yarn build` locally too, maybe we can spot the issue there.

karelianpie:
> > @Majorfi I've asked in YFI Infrastructure why the AMS Amplify is failing. Will let you know
> 
> Lets try running `yarn build` locally too, maybe we can spot the issue there.

```cmd
➜  yearn-finance-v3 git:(web-lib) yarn build
yarn run v1.22.18
$ cross-env REACT_APP_VERSION=$(git rev-parse HEAD) craco build
The following changes are being made to your tsconfig.json file:
  - compilerOptions.paths must not be set (aliased imports are not supported)

Creating an optimized production build...
Failed to compile.

./src/index.tsx
Cannot find module: '@yearn/web-lib/dist/build.css'. Make sure this package is installed.

You can install this package by running: yarn add @yearn/web-lib/dist/build.css.


error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
then when I run `yarn` I get the following

```cmd
➜  yearn-finance-v3 git:(web-lib) yarn
yarn install v1.22.18
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "https://npm.pkg.github.com/download/@yearn/web-lib/0.2.27/a5aec6437505213aef6fb84cf564053246dd21b7cb0105c33c517bfe872ed3bc: Request failed \"401 Unauthorized\"".
info If you think this is a bug, please open a bug report with the information provided in "/Users/yearn/yearn-finance-v3/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

I remember we've encountered something similar when doing the web-lib-ui

.github/workflows/nodejs.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v3
-        with:
```
karelianpie, line 29: I'm confused, what happened here for this to build? Did you do something in AWS Amplify?

xgambitox: yeah, would be nice to know what was the issue, so we can avoid it in the future.

Will approve since all looks good and build working.

xgambitox (approved):

karelianpie (approved):
LGTM

.github/workflows/nodejs.yml
mariuspod: @karelianpie the amplify builder needs to have access to the github npm package registry for installing libs.
I've added this line to our amplify build settings to make the build succeed (needs to be done on all repos) :
`- echo "//npm.pkg.github.com/:_authToken=${GITHUB_NPM_TOKEN}" >> .npmrc`

karelianpie: Thanks @mariuspod! That makes sense 🙇

PR #625, opened 2022-05-02
feat: ledger live customizations

xgambitox:
## Description

- Add partner state to redux store
- When app loads as ledger live app:
  - Hide network selection (only Mainnet support)
  - Wallet select functionality disabled (injected by ledger)
  - Hide Labs nav link
  - Hide Iron Bank nav link
  - Hide and disable signed approvals setting

## Motivation and Context

- Requested changes by ledger

## How Has This Been Tested?

Tested UI changes locally by changing variables. Will require further testing on a ledger live dev environment.


src/core/store/modules/app/app.actions.ts
```diff
@@ -34,11 +36,16 @@ const clearUserAppData = createAsyncThunk<void, void, ThunkAPI>('app/clearUserAp
 /*                                 Fetch Data                                 */
 /* -------------------------------------------------------------------------- */
 
-const initApp = createAsyncThunk<void, void, ThunkAPI>('app/initApp', async (_arg, { dispatch, getState }) => {
+const initApp = createAsyncThunk<void, void, ThunkAPI>('app/initApp', async (_arg, { dispatch, getState, extra }) => {
+  const { CONTRACT_ADDRESSES } = extra.config;
   const { wallet, network } = getState();
-  if (inIframe()) {
+  if (inLedgerIframe()) {
+    if (network.current !== 'mainnet') await dispatch(NetworkActions.changeNetwork({ network: 'mainnet' }));
```
turtlemoji, line 43: we had problems in the past forcing to change networks, since some wallets doesn't like this behaviour, to have it in mind

src/core/store/modules/app/app.actions.ts
```diff
@@ -34,11 +36,16 @@ const clearUserAppData = createAsyncThunk<void, void, ThunkAPI>('app/clearUserAp
...
     await dispatch(WalletActions.walletSelect({ walletName: 'Iframe', network: 'mainnet' }));
-  } else if (wallet.name) {
+    await dispatch(PartnerActions.changePartner({ partnerId: 'ledger', address: CONTRACT_ADDRESSES.LEDGER }));
+  } else if (inIframe()) {
+    if (network.current !== 'mainnet') await dispatch(NetworkActions.changeNetwork({ network: 'mainnet' }));
```
turtlemoji, line 47: idem

turtlemoji (approved):

src/core/store/modules/app/app.actions.ts
xgambitox: Yeah, this part changes the selected network for the app, and it only dispatch an event for the wallet network change if its Metamask now. This part its just meant if for some reason ledger live loads in another network (which should not happen), since they just support mainnet.

src/core/store/modules/app/app.actions.ts
xgambitox: We can test this part on Gnosis when it gets merged, its just a fallback if for any reason its not set to the supported network.

src/core/types/State.ts
```diff
@@ -241,3 +243,8 @@ export interface UserState {
     getNftBalance: Status;
   };
 }
+
+export interface PartnerState {
+  id: PartnerId | undefined;
+  address: Address | undefined;
```
karelianpie, line 249:
```suggestion
  id?: PartnerId;
  address?: Address;
```

src/core/store/modules/partner/partner.reducer.ts
```diff
@@ -0,0 +1,21 @@
+import { createReducer } from '@reduxjs/toolkit';
+
+import { PartnerState } from '@types';
+
+import { PartnerActions } from './partner.actions';
+
+export const partnerInitialState: PartnerState = {
+  id: undefined,
+  address: undefined,
+};
```
karelianpie, line 10: This doesn't seem needed, how about we just do the following below
```typescript
createReducer({}, (builder) => {})
```

src/core/types/Partner.ts
```diff
@@ -0,0 +1 @@
+export type PartnerId = 'ledger';
```
karelianpie, line 1: Is the intention here to have multiple partner ids?

src/core/store/modules/app/app.actions.ts
```diff
@@ -34,11 +36,16 @@ const clearUserAppData = createAsyncThunk<void, void, ThunkAPI>('app/clearUserAp
 /*                                 Fetch Data                                 */
 /* -------------------------------------------------------------------------- */
 
-const initApp = createAsyncThunk<void, void, ThunkAPI>('app/initApp', async (_arg, { dispatch, getState }) => {
+const initApp = createAsyncThunk<void, void, ThunkAPI>('app/initApp', async (_arg, { dispatch, getState, extra }) => {
+  const { CONTRACT_ADDRESSES } = extra.config;
   const { wallet, network } = getState();
```
karelianpie, line 41: nit: Wondering if we should have some checks here for `wallet` and `network`.

Also, it's a bit nested, could we refactor this so we just return when we're done instead of having if/else?

src/core/store/modules/app/app.actions.ts
```diff
@@ -34,11 +36,16 @@ const clearUserAppData = createAsyncThunk<void, void, ThunkAPI>('app/clearUserAp
...
-  if (inIframe()) {
+  if (inLedgerIframe()) {
+    if (network.current !== 'mainnet') await dispatch(NetworkActions.changeNetwork({ network: 'mainnet' }));
     await dispatch(WalletActions.walletSelect({ walletName: 'Iframe', network: 'mainnet' }));
-  } else if (wallet.name) {
+    await dispatch(PartnerActions.changePartner({ partnerId: 'ledger', address: CONTRACT_ADDRESSES.LEDGER }));
```
karelianpie, line 45: This logic could even live in a `handleInLedgerIFrame()`

src/core/store/modules/partner/partner.actions.ts
```diff
@@ -0,0 +1,9 @@
+import { createAction } from '@reduxjs/toolkit';
+
+import { Address, PartnerId } from '@types';
+
+const changePartner = createAction<{ partnerId: PartnerId; address: Address }>('partner/changePartner');
```
karelianpie, line 5: In here we have `partnerId: PartnerId; address: Address` but then in `PartnerState` we have  `id: PartnerId; address: Address;` -- should we make this consistent and go with `id` or `partnerId` but not both?

src/client/components/app/Navigation/index.tsx
```diff
@@ -48,8 +51,13 @@ const navLinks: NavigationLink[] = [
...
 ];
 
-export const Navigation = () => {
+interface NavigationProps {
+  hideOptionals?: boolean;
+}
+
+export const Navigation = ({ hideOptionals }: NavigationProps) => {
   const { isMobile, isTablet, isDesktop } = useWindowDimensions();
+  const displayLinks = navLinks.filter((link) => !(link.optional && hideOptionals));
```
karelianpie, line 60: I find `!(link.optional && hideOptionals))` a bit hard to wrap my head around, `hideOptionals` seems a bit mysterious. What optionals? Maybe renaming it?

src/core/types/State.ts
xgambitox: We set the state types intentionally not optional just so that we have to explicitly define the whole initial state on the reducer and all its properties, for readability and avoiding errors.

I would agree with changing to `null`, but would rather make that refactor for all state in this file in a separate PR, just so we have a convention, as it will be hard to know for other contributors which one to use.

Will leave it as is for now since it does not affect functionality.

src/core/store/modules/partner/partner.reducer.ts
xgambitox: Yeah, agree with you here on better using `null`. We can transition to this approach in a refactor for all the state so we have a convention.

Will leave as is for now.

src/core/types/Partner.ts
xgambitox: yes, eventually we will add more partners here.

src/core/store/modules/app/app.actions.ts
xgambitox: Since `wallet` and `network` have an initial state set by the reducer when the store gets created, dont think checks are necessary for wallet and network here at init.

Yes it does have some nested `if` conditions, but could not think of a better way to handle them :P

src/core/store/modules/app/app.actions.ts
xgambitox: I would agree to move part of the code to another thunk if we have a need to reuse the function. For now it seems fine to me, since it would only be called here.

src/core/store/modules/partner/partner.actions.ts
xgambitox: I left the state as `id` just because when you access the property from the state object, its better imo to have `partner.id` rather that `partner.partnerId`. 

how about changing it to:
`const changePartner = createAction<{ id: PartnerId; address: Address }>('partner/changePartner');`
?

Other thing we can consider is having an interface for the props, like
```
interface changePartnerProps {
   id: PartnerId;
   address: Address;
}
```
I wouldnt use the `PartnerState` here just because it might change in the future, and have more properties.

src/core/store/modules/app/app.actions.ts
karelianpie: All good, maybe we can refactor later, I have some ideas

src/core/store/modules/partner/partner.actions.ts
karelianpie: Sounds good, I'd just make the interface camelCase 

```typescript
interface ChangePartnerProps {
   id: PartnerId;
   address: Address;
}
```

so then we can use

```typescript
const changePartner = createAction<ChangePartnerProps>('partner/changePartner');
```

src/client/containers/Layout/index.tsx
```diff
@@ -189,9 +194,12 @@ export const Layout: FC = ({ children }) => {
           walletAddress={selectedAddress}
           addressEnsName={addressEnsName}
           onWalletClick={() => dispatch(WalletActions.walletSelect({ network: currentNetwork }))}
+          disableWalletSelect={hideControls || isInIframe}
```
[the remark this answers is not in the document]
xgambitox: Right now we also have the integration of our site as a Gnosis App that runs as an Iframe. Since it is not really considered a partner, left the iFrame checks there too.

src/core/store/modules/app/app.actions.ts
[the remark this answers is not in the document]
xgambitox: It changes the app selected network, and only triggers the prompt if its a MM wallet. Since here we know its being loaded in Ledger Live, it will not trigger the prompt.

src/core/store/modules/partner/partner.actions.ts
xgambitox: You are right, made a typo there, ty.

xgambitox: added here https://github.com/yearn/yearn-finance-v3/pull/625/commits/399b6b089ac3b90f2dcce07b0e58f594a04974b0

src/client/components/app/Navigation/index.tsx
xgambitox: changed here https://github.com/yearn/yearn-finance-v3/pull/625/commits/808e3aba91bb9f276eafc615ee8bad5594870553

src/core/frameworks/yearnSdk/index.ts
```diff
@@ -16,7 +16,7 @@ export class YearnSdkImpl implements YearnSdk {
       const networkId = getNetworkId(network) as SdkNetwork;
       const sdkInstance = new Yearn(networkId, {
         provider,
-        partnerId: isLedger && networkId === 1 ? CONTRACT_ADDRESSES.LEDGER_PARTNER_ID : undefined,
+        partnerId: isLedger && networkId === 1 ? CONTRACT_ADDRESSES.LEDGER : undefined,
```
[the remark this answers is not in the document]
xgambitox: changed it back to original here https://github.com/yearn/yearn-finance-v3/pull/625/commits/5c9e29ccf3b175ff8d50217226dda013842a3a73 :)

turtlemoji (approved):

karelianpie (approved):

PR #628, opened 2022-05-03
fix: portfolio wallet deposit modal vault select

### LINKED ISSUE #626: fix: portfolio wallet deposit modal
**Describe the bug**
From the portfolio page, cannot change which vault depositing into

**To Reproduce**
1. Go to portfolio page
2. Click on deposit from any token in wallet

**Expected behavior**
Should be able to change which vault being deposited or zapped into

**Screenshots**
wbtc can only go to wbtc from wallet 
image

but from other vaults I can deposit wbtc

CRV is only pointing to rocket (like many other vaults, rocket is default)
image

but from vault details on other vaults I can zap.

**Environment Information**

**Additional context**

edwinzhng:
## Description

* Currently, the transaction deposit modal expects the URL path to be `/wallet` when it sets the `allowVaultSelect` parameter `true`, this doesn't happen right now since the new page URL path is `/portfolio`
* I updated the code to pass in props to the modal instead of reading the path, so that this is more explicit and shouldn't happen again if we ever change the page URL path later down the road

## Related Issue

Fixes #626 

## Motivation and Context

Now users can select a vault to zap into again instead of just the default

## How Has This Been Tested?

Manually tested on portfolio wallet page with different tokens. Also tested the vault page and confirmed that functionality is still the same there (can change token but cannot change vault to deposit into).

Screen Shot 2022-05-03 at 5 11 08 PM

Screen Shot 2022-05-03 at 5 13 19 PM

### DIFF
--- a/src/client/routes/Portfolio/index.tsx
+++ b/src/client/routes/Portfolio/index.tsx
@@ -141,7 +141,12 @@ export const Portfolio = () => {
     switch (action) {
       case 'invest':
         dispatch(TokensActions.setSelectedTokenAddress({ tokenAddress }));
-        dispatch(ModalsActions.openModal({ modalName: 'depositTx' }));
+        dispatch(
+          ModalsActions.openModal({
+            modalName: 'depositTx',
+            modalProps: { allowTokenSelect: false, allowVaultSelect: true },
+          })
+        );
         break;
       default:
         break;
--- a/src/client/containers/Modals/DepositTxModal/index.tsx
+++ b/src/client/containers/Modals/DepositTxModal/index.tsx
@@ -1,6 +1,5 @@
 import { FC } from 'react';
 import styled from 'styled-components';
-import { useLocation } from 'react-router-dom';
 
 import { useAppTranslation } from '@hooks';
 import { ModalTx } from '@components/common';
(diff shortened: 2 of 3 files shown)


src/client/routes/Portfolio/index.tsx
```diff
@@ -141,7 +141,12 @@ export const Portfolio = () => {
     switch (action) {
       case 'invest':
         dispatch(TokensActions.setSelectedTokenAddress({ tokenAddress }));
-        dispatch(ModalsActions.openModal({ modalName: 'depositTx' }));
+        dispatch(
+          ModalsActions.openModal({
+            modalName: 'depositTx',
+            modalProps: { allowTokenSelect: false, allowVaultSelect: true },
```
karelianpie, line 147: Just wondering if this is the only place we want `{ allowTokenSelect: false, allowVaultSelect: true }`

karelianpie (approved):
LGTM

src/client/routes/Portfolio/index.tsx
edwinzhng: I think so, since previously it was hard-coded to only do that on the "/wallet" path. Couldn't find anywhere else we load this modal other than the vaults page

FoxTheSin:
possibly need to apply a different fix before this?
image

turtlemoji (approved):

src/client/routes/Portfolio/index.tsx
karelianpie: Nice! All good

xgambitox (approved):

xgambitox:
> possibly need to apply a different fix before this? image

Yeah, this was an issue with the allowlist and not related to this fix.

Good one @edwinzhng :)

PR #630, opened 2022-05-04
fix: deposit token allowance

xgambitox:
## Description

Fixes a race condition on vaults detail view, in which the wallet is not yet connected before it executes, making users unable to deposit.
image

## Motivation and Context

Refactored the deposit modals to fetch token allowance directly instead of using useAllowance hook, since was unable to make it work through the hook.

## How Has This Been Tested?

Reloaded vault detail page multiple times with wallet connected.
Approved tokens in fantom network to test functionality. Also tested labs.
@FoxTheSin help needed with QA.

### DIFF
--- a/src/client/components/app/Transactions/DepositTx.tsx
+++ b/src/client/components/app/Transactions/DepositTx.tsx
@@ -56,7 +56,7 @@ export const DepositTx: FC<DepositTxProps> = ({
   const [isFetchingAllowance, setIsFetchingAllowance] = useState(false);
   const currentNetwork = useAppSelector(NetworkSelectors.selectCurrentNetwork);
   const walletNetwork = useAppSelector(WalletSelectors.selectWalletNetwork);
-  const walletIsConnected = useAppSelector(WalletSelectors.selectWalletIsConnected);
+  const isWalletConnected = useAppSelector(WalletSelectors.selectWalletIsConnected);
   const currentNetworkSettings = NETWORK_SETTINGS[currentNetwork];
   const vaults = useAppSelector(VaultsSelectors.selectLiveVaults);
   const selectedVault = useAppSelector(VaultsSelectors.selectSelectedVault);
@@ -107,7 +107,7 @@ export const DepositTx: FC<DepositTxProps> = ({
   }, []);
 
   useEffect(() => {
-    if (!selectedVault || !selectedSellTokenAddress || !walletIsConnected) return;
+    if (!selectedVault || !selectedSellTokenAddress || !isWalletConnected) return;
     const fetchAllowance = async () => {
       setIsFetchingAllowance(true);
       await dispatch(
@@ -119,7 +119,7 @@ export const DepositTx: FC<DepositTxProps> = ({
       setIsFetchingAllowance(false);
     };
     fetchAllowance();
-  }, [selectedSellTokenAddress, selectedVault?.address, walletIsConnected]);
+  }, [selectedSellTokenAddress, selectedVault?.address, isWalletConnected]);
 
   useEffect(() => {
     if (!selectedVault) return;
--- a/src/client/components/app/Transactions/DepositTx.tsx
+++ b/src/client/components/app/Transactions/DepositTx.tsx
@@ -111,7 +111,7 @@ export const DepositTx: FC<DepositTxProps> = ({
     const fetchAllowance = async () => {
       setIsFetchingAllowance(true);
       await dispatch(
-        VaultsActions.getVaultAllowance({
+        VaultsActions.getDepositAllowance({
           vaultAddress: selectedVault.address,
           tokenAddress: selectedSellTokenAddress,
         })
--- a/src/client/components/app/Transactions/DepositTx.tsx
+++ b/src/client/components/app/Transactions/DepositTx.tsx
@@ -6,7 +6,6 @@ import {
   useAppDispatchAndUnwrap,
   useDebounce,
   useAppTranslation,
-  useAllowance,
   useSelectedSellToken,
 } from '@hooks';
 import {
@@ -27,7 +26,7 @@ import {
   validateSlippage,
   validateNetwork,
   formatApy,
-  basicValidateAllowance,
+  validateVaultAllowance,
 } from '@utils';
 import { getConfig } from '@config';
 
@@ -54,8 +53,10 @@ export const DepositTx: FC<DepositTxProps> = ({
   const [amount, setAmount] = useState('');
   const [debouncedAmount, isDebouncePending] = useDebounce(amount, 500);
   const [txCompleted, setTxCompleted] = useState(false);
+  const [isFetchingAllowance, setIsFetchingAllowance] = useState(false);
   const currentNetwork = useAppSelector(NetworkSelectors.selectCurrentNetwork);
   const walletNetwork = useAppSelector(WalletSelectors.selectWalletNetwork);
+  const walletIsConnected = useAppSelector(WalletSelectors.selectWalletIsConnected);
   const currentNetworkSettings = NETWORK_SETTINGS[currentNetwork];
   const vaults = useAppSelector(VaultsSelectors.selectLiveVaults);
   const selectedVault = useAppSelector(VaultsSelectors.selectSelectedVault);
@@ -68,10 +69,6 @@ export const DepositTx: FC<DepositTxProps> = ({
     selectedSellTokenAddress,
     selectedVaultOrLab: selectedVault,
     allowTokenSelect,
-  });
-  const [tokenAllowance, isLoadingTokenAllowance, tokenAllowanceError] = useAllowance({
-    tokenAddress: selectedSellTokenAddress,
-    vaultAddress: selectedVault?.address,
   });
 
   const onExit = () => {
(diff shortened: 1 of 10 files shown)


src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -199,7 +198,7 @@ const approveDeposit = createAsyncThunk<
       vaultAddress: vaultData.address,
     });
 
-    await transactionService.handleTransaction({ tx: tx as TransactionResponse, network: network.current });
+    await transactionService.handleTransaction({ tx, network: network.current });
```
karelianpie, line 201: 👍 for removing the casting

src/core/store/modules/vaults/vaults.actions.ts
```diff
@@ -456,12 +455,22 @@ const getVaultAllowance = createAsyncThunk<
 
   if (!userAddress) throw new Error('WALLET NOT CONNECTED');
 
-  return vaultService.getVaultAllowance({
+  const tokenAllowance = await vaultService.getVaultAllowance({
```
karelianpie, line 458: I feel this sounds confusing, we are assigning the result of `getVaultAllowance()` to `tokenAllowance`?

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
```diff
@@ -149,9 +157,8 @@ export const LabDepositTx: FC<LabDepositTxProps> = ({ onClose }) => {
       expectedTxOutcomeStatus.error ||
       actionsStatus.approveDeposit.error ||
       actionsStatus.deposit.error ||
-      slippageError ||
-      tokenAllowanceError,
-    loading: expectedTxOutcomeStatus.loading || isDebouncePending || isLoadingTokenAllowance,
+      slippageError,
+    loading: expectedTxOutcomeStatus.loading || isDebouncePending,
```
karelianpie, line 161: When `isFetchingAllowance` is not loading for the target status?

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
```diff
@@ -110,17 +121,14 @@ export const LabDepositTx: FC<LabDepositTxProps> = ({ onClose }) => {
...
 
-  let isApproved: boolean | undefined;
-  let allowanceError: string | undefined;
-
-  const { approved, error } = basicValidateAllowance({
-    tokenAddress: selectedSellTokenAddress,
-    tokenAmount: toBN(debouncedAmount),
-    tokenDecimals: selectedSellToken.decimals.toString(),
-    rawAllowance: tokenAllowance?.amount || '0',
+  const { approved: isApproved, error: allowanceError } = validateVaultAllowance({
```
karelianpie, line 124: How about here just changing the return from `validateVaultAllowance()` as it appears we were not and still aren't happy with the naming?

src/core/store/modules/vaults/vaults.actions.ts
xgambitox: used that name since `TokenAllowance` is the type that is returned by this method. It is indeed a tokenAllowance. The service should be called `getVaultDepositAllowance` though imo. Since we will later add a different one for withdrawals.

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
xgambitox: it just disables the approve button when it is loading now, imo it should not affect the simulation output from showing, since either it is approved or not, but target expected outcome persists anyway

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
xgambitox: not following the issue here. can you explain? not sure if we should rename the validation function name

src/core/store/modules/vaults/vaults.actions.ts
karelianpie: So not `getTokenAllowance()`?

src/client/components/app/Transactions/DepositTx.tsx
```diff
@@ -54,8 +53,10 @@ export const DepositTx: FC<DepositTxProps> = ({
   const [amount, setAmount] = useState('');
   const [debouncedAmount, isDebouncePending] = useDebounce(amount, 500);
   const [txCompleted, setTxCompleted] = useState(false);
+  const [isFetchingAllowance, setIsFetchingAllowance] = useState(false);
   const currentNetwork = useAppSelector(NetworkSelectors.selectCurrentNetwork);
   const walletNetwork = useAppSelector(WalletSelectors.selectWalletNetwork);
+  const walletIsConnected = useAppSelector(WalletSelectors.selectWalletIsConnected);
```
karelianpie, line 59: nit: Usually the best practice is to just have these like below
```suggestion
  const isWalletConnected = useAppSelector(WalletSelectors.selectWalletIsConnected);
```
No action needed, but would be quicker to see that that returns a boolean

src/client/components/app/Transactions/DepositTx.tsx
```diff
@@ -109,6 +106,21 @@ export const DepositTx: FC<DepositTxProps> = ({
...
+      await dispatch(
+        VaultsActions.getVaultAllowance({
+          vaultAddress: selectedVault.address,
+          tokenAddress: selectedSellTokenAddress,
+        })
+      );
+      setIsFetchingAllowance(false);
+    };
+    fetchAllowance();
+  }, [selectedSellTokenAddress, selectedVault?.address, walletIsConnected]);
```
karelianpie, line 122: Would it be possible to extract just this bit to its own hook so we can reuse in `LabDepositTx.tsx` -- if too hard let's just leave as is

FoxTheSin:
So I wasn't sure how to test this so I just did an approve in labs (not sure if relevant to this issue though)
(vbnt to yvBOOST)
Approval window was stuck open (despite tx going through)
image

I then refreshed to try deposit, this happened
image

src/core/store/modules/vaults/vaults.actions.ts
xgambitox: I would rename it to `getDepositAllowance` since its implicit it refers to vaults since its in the `vaultService`. That way its also aligned with sdk here https://github.com/yearn/yearn-sdk/blob/master/src/interfaces/vault.ts#L337. wdyt?

src/client/components/app/Transactions/DepositTx.tsx
xgambitox: Yeah, agree with this, just used the same naming from the selector, which probably should also be named as you describe

src/client/components/app/Transactions/DepositTx.tsx
xgambitox: Yeah, it would be ideal. Should we refactor in a new PR, so we release a fix for this bug that is in prod now?

xgambitox:
> So I wasn't sure how to test this so I just did an approve in labs (not sure if relevant to this issue though) (vbnt to yvBOOST) Approval window was stuck open (despite tx going through) image
> 
> refreshed page after confirming "approval", and before clicking deposit, simulation appeared to pass image
> 
> I then I clicked deposit and this happened image

the transaction notification never works on preview builds since the generated URL is not whitelisted in blocknatives dashboard.

About the error from vBNT -> yvboost, not sure if its related to this PR. Can you try in prod if same thing happens? That is how we can discard it from being related to this one.

I would do QA with more liquid tokens though, since we might get other type of error non related to this PR if not.

FoxTheSin:
> 
> the transaction notification never works on preview builds since the generated URL is not whitelisted in blocknatives dashboard.
> 
> About the error from vBNT -> yvboost, not sure if its related to this PR. Can you try in prod if same thing happens? That is how we can discard it from being related to this one.
> 
> I would recommend doing QA with more liquid tokens though, since we might get other type of error non related to this PR if not.

Yup, doesn't work in prod.
but everything else looked good!

FoxTheSin:
another weird occurrence, 
1) opened the amplify for pr (wallet was set to eth)
2) switched to ftm from meta mask (in portfolio page)
3) changed to vaults page
4) see below! (list of eth and ftm got combined.)

image

xgambitox:
> another weird occurrence (but unable to reproduce)
> 
> 1. opened the amplify for pr (wallet was set to eth)
> 2. switched to ftm from meta mask (in portfolio page)
> 3. changed to vaults page
> 4. see below! (list of eth and ftm got combined.)
> 
> image

Yeah, I have seen this one, and might be good to track on another issue. It seems related to fetching been slow, and therefore vaults state not getting cleared properly

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
karelianpie: Sorry I wasn't very clear, what I meant was to change the returned object keys, so instead of `approve` we would return `isApproved` from the `validateVaultAllowance()` as it seems that's the name we want and it's best practice when returning a boolean to prefix with "is"

src/client/components/app/Transactions/DepositTx.tsx
karelianpie: Sounds good

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
karelianpie: Let's leave as is for now so we can resolve the bug and maybe then get back to this later. Resolving

karelianpie (approved):
LGTM

src/core/store/modules/vaults/vaults.actions.ts
karelianpie: That sounds good to me!

turtlemoji (dismissed):

turtlemoji:
@xgambitox maybe bump package so we can just release

src/core/store/modules/vaults/vaults.actions.ts
xgambitox: refactored here :) https://github.com/yearn/yearn-finance-v3/pull/630/commits/e73ba66db30b3b4143aac516d62beae4ec40176b

src/client/components/app/Transactions/DepositTx.tsx
xgambitox: changed here https://github.com/yearn/yearn-finance-v3/pull/630/commits/6f98c168328fc60e05d693fbaadc3c2a0a309621

xgambitox:
@FoxTheSin @karelianpie @turtlemoji should be ready now 👍

turtlemoji (approved):

karelianpie (approved):

PR #634, opened 2022-05-06
fix: pull yvBOOST balance from lab positions and fix icon URL

### LINKED ISSUE #627: fix: deposit to pickle
**Describe the bug**
when in deposit modal (for pickle jar) it isn't recognizing yvBOOST balance

**To Reproduce**
1. Go to labs
2. select deposit
3. select yvBOOST as deposit token

**Expected behavior**
Should be able to zap in yvBOOST from deposit modal

**Screenshots**
0 balance listed when balance is available to deposit.
image

**Environment Information**

**Additional context**

edwinzhng:
## Description

* It seems like yvBOOST token balances don't show up for pickle jar because the yvBOOST token is not fetched with all other user tokens. To fix this, I override the user tokens balance with the lab positions data. This is kind of hacky but it seems cleaner than trying to force fetch yvBOOST tokens with all the other user tokens

* There was also an issue with a broken asset icon URL in TokenService (but the correct URL was in LabService), so I fixed that by creating a constant to share between them

## Related Issue

Fixes #627

## Motivation and Context

yvBOOST token balances now show up for pickle jar

## How Has This Been Tested?

Manually checked labs with the following address (which has assets in pickle jar): 0x24a21d91c82d684cd793e3829318a33784ef3626

Screen Shot 2022-05-06 at 1 01 44 PM

Screen Shot 2022-05-06 at 1 01 58 PM

### DIFF
--- a/src/core/services/LabService.ts
+++ b/src/core/services/LabService.ts
@@ -65,6 +65,7 @@ export class LabServiceImpl implements LabService {
     const errors: string[] = [];
     const { YEARN_API, CONTRACT_ADDRESSES } = this.config;
     const { ETH, YVECRV, CRV, YVBOOST, PSLPYVBOOSTETH } = CONTRACT_ADDRESSES;
+    const { ASSETS_ICON_URL } = getConstants();
     const yearn = this.yearnSdk.getInstanceOf(network);
     const providerType = getProviderType(network);
     const provider = this.web3Provider.getInstanceOf(providerType);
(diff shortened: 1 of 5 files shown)


FoxTheSin:
tested and appears working!
image

src/core/services/LabService.ts
```diff
@@ -129,7 +128,7 @@ export class LabServiceImpl implements LabService {
               }
             : undefined,
           displayName: backscratcherData.name,
-          displayIcon: `${ASSET_URL}${YVECRV}/logo-128.png`,
+          displayIcon: `${getConstants().ASSETS_ICON_URL}${YVECRV}/logo-128.png`,
```
xgambitox, line 131: nit: how about destructuring the object to access the property needed from `getConstants()`, since its used multiple times?

xgambitox (dismissed):
Solution looks good to me. These Labs section can get hacky since its very custom, so its fine for now. Good one 👍

xgambitox (approved):

turtlemoji (approved):

turtlemoji:
nice job @edwinzhng ty

PR #635, opened 2022-05-07
fix: Improve visibility of icon button color

### LINKED ISSUE #629: Available to stake action button icon not visible and undescriptive
**Describe the bug**

"Available to stake" ActionButtons > AlertButton has a white background which makes the SVG not visible

Also, there's no tooltip to explain what the issue is, as when hovering only the title is displayed (suggestion: "Stake needed to earn APY")

telegram-cloud-photo-size-5-6309576523312639702-x

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to settings and use an account that has pSLPyvBOOST-ETH
2. Go to labs and in the Holdings section observe the issue

**Expected behaviour**

See the icon and a more descriptive issue when hovering the warning

**Screenshots**

telegram-cloud-document-5-6309576522856400186

When the background colour is changed it becomes visible

telegram-cloud-photo-size-5-6309576523312639700-x

huggingbot:
## Description

- Icon button is not clearly visible for certain themes, e.g. classic.
- Added `colors.vaultActionButton.fillColor` prop to use `colors.backgroundVariant` which makes icon button visible for all themes.

## Related Issue
#629

## Motivation and Context

- Improve visibility of icon button and user experience.

## How Has This Been Tested?

- Running locally and checking icon color in all the themes.

- Refer to #629.

### DIFF
--- a/src/client/components/app/ActionButtons/index.tsx
+++ b/src/client/components/app/ActionButtons/index.tsx
@@ -20,7 +20,7 @@ const AlertButton = styled.div`
   justify-content: center;
   border-radius: 0.8rem;
   border: 2px solid ${({ theme }) => theme.colors.vaultActionButton.selected.borderColor};
-  background: ${({ theme }) => theme.colors.vaultActionButton.borderColor};
+  background: ${({ theme }) => theme.colors.vaultActionButton.iconFill};
   width: 2.8rem;
   height: 2.8rem;
 `;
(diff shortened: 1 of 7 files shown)


turtlemoji:
amazing ty!

turtlemoji (approved):

karelianpie (approved):
LGTM, thanks @huggingbot! 

One small thing, I don't think we can have the "Partially fixes" otherwise it will close the issue even though we still don't have the hovering text. I'd be happy adding that to this PR -- or if you would like to take that as well then we should either remove that text from this PR's description so the issue is not closed or I can create another issue for just the hovering text. Let me know how you'd like to proceed

karelianpie:
> Thanks @karelianpie for the heads-up! Have removed the closing text. Actually tried to update the text but didn't have the translation for all the languages. Would be good if we could address that in a separate PR.

Sounds good 👍

PR #642, opened 2022-05-09
fix: gnosis app wallet connect

xgambitox:
## Description

Fix issue with wallet not connecting when loaded as a gnosis safe app 

## How Has This Been Tested?

We will need to test it in gnosis once it gets deployed to prod, since I could not add the preview build as a custom app in gnosis safe.

### DIFF
--- a/src/utils/misc.ts
+++ b/src/utils/misc.ts
@@ -58,10 +58,7 @@ export const inIframe = () => {
 
 export const inLedgerIframe = () => {
   try {
-    return (
-      window.self !== window.top &&
-      (window?.location?.ancestorOrigins?.[0]?.includes('ledger') || window?.navigator?.userAgent.includes('Firefox'))
-    );
+    return window.self !== window.top && window?.location?.ancestorOrigins?.[0]?.includes('ledger');
   } catch (e) {
     return true;
   }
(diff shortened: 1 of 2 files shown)


src/utils/misc.ts
```diff
@@ -58,10 +58,7 @@ export const inIframe = () => {
 
 export const inLedgerIframe = () => {
   try {
-    return (
-      window.self !== window.top &&
-      (window?.location?.ancestorOrigins?.[0]?.includes('ledger') || window?.navigator?.userAgent.includes('Firefox'))
-    );
+    return window.self !== window.top && window?.location?.ancestorOrigins?.[0]?.includes('ledger');
```
turtlemoji, line 61: maybe we can try document.referrer for firefox?

turtlemoji: we need to check the string value on ledger and gnosis to see if we can work with it tho
https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

turtlemoji: will be empty if is not an iframe

turtlemoji: also maybe check window.location.origin

xgambitox: Since ledger live is an installed app in the OS, and not really loaded through a browser, can we assume this condition is not needed?

turtlemoji (approved):

FoxTheSin (approved):

PR #647, opened 2022-05-10
feat: add breadcrumb to vaults for vault details

### LINKED ISSUE #646: feat: vaults breadcrumb to link to vaults page
**Elevator Pitch**
The main header (H1) should have a link back to vault when in the vault details.

**Customer Value**
Allow users to easily navigate to vaults.

**Definition of Done**
This is linked to vaults when in vault details
image

**Design Approach**
Approach and considerations to build the task list / Existing code patterns in production code base that you will base your work off of.

**Additional Context**
Screenshots, UI/UX designs, links or other that will help implementing the feature.

edwinzhng:
## Description

* Adds breadcrumbs that link back to vault page

## Related Issue

Fixes #646

## How Has This Been Tested?

* Manually tested clicking the link, keyboard navigation

Screen Shot 2022-05-10 at 8 28 28 AM

### DIFF
--- a/src/client/containers/Layout/index.tsx
+++ b/src/client/containers/Layout/index.tsx
@@ -107,9 +107,11 @@ export const Layout: FC = ({ children }) => {
   const hideControls = partner.id === 'ledger';
 
   let vaultName;
+  let titleLink;
   // TODO Add lab details route when its added the view
   if (path === 'vault') {
     vaultName = selectedVault?.displayName;
+    titleLink = '/vaults';
   }
 
   // TODO This is only assetAddress on the vault page
@@ -188,6 +190,7 @@ export const Layout: FC = ({ children }) => {
       <Content collapsedSidebar={collapsedSidebar} useTabbar={isMobile}>
         <Navbar
           title={t(`navigation.${path}`)}
+          titleLink={titleLink}
           subTitle={vaultName}
           walletAddress={selectedAddress}
           addressEnsName={addressEnsName}
--- a/src/client/components/app/Navbar/index.tsx
+++ b/src/client/components/app/Navbar/index.tsx
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
 
 import { ConnectWalletButton } from '@components/app';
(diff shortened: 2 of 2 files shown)


src/client/containers/Layout/index.tsx
```diff
@@ -107,9 +107,11 @@ export const Layout: FC = ({ children }) => {
   const hideControls = partner.id === 'ledger';
 
   let vaultName;
+  let titleLink;
   // TODO Add lab details route when its added the view
   if (path === 'vault') {
     vaultName = selectedVault?.displayName;
+    titleLink = '/vaults';
```
edwinzhng, line 114: I'm thinking that maybe all of these routes should be pulled out into a constant so that we don't need to hard code them everywhere? The `'/vaults'` route for example is used in 3 separate locations now, any thoughts? Should probably do this in a different PR though

turtlemoji: yeah, the most approximate think is the navigation const navLinks
```
const navLinks: NavigationLink[] = [
  {
    to: '/portfolio',
    text: 'navigation.portfolio',
    icon: WalletIcon,
  },
  {
    to: '/vaults',
    text: 'navigation.vaults',
    icon: VaultIcon,
  },
  {
    to: '/labs',
    text: 'navigation.labs',
    icon: LabsIcon,
    optional: true,
  },
  {
    to: 'https://app.ib.xyz',
    text: 'navigation.ironbank',
    icon: IronBankIcon,
    external: true,
    optional: true,
  },
  {
    to: '/settings',
    text: 'navigation.settings',
    icon: SettingsIcon,
    // hideMobile: true,
  },
];
```

turtlemoji (approved):

src/client/containers/Layout/index.tsx
xgambitox: yeah agree @edwinzhng. Since right now we only have one nested route, we can just merge this one as is, with the intention of refactor it in the future since it will most likely grow.

xgambitox (approved):

PR #651, opened 2022-05-11
refactor: Update zapper's endpoints

karelianpie:
## Description

Refactor to use the new zapper v2 endpoints

## Related Issue

https://medium.com/zapper-protocol/updates-to-zappers-public-apis-4933a2967f57

## Motivation and Context

Old endpoints have been deprecated and will be sunset on May 25th, 2022

They've also changed the way we pass the api key, before as a url param and now in the header

## How Has This Been Tested?

* Hit the v2 endpoint to make sure the result is the same as the v1
* Ran the FE locally with the new endpoints
* Added a test for the auth token encoding `src/utils/encode.spec.ts`


turtlemoji (dismissed):

src/config/env/index.ts
```diff
@@ -17,7 +18,7 @@ export const getEnv = memoize(
     BLOCKNATIVE_KEY: process.env.REACT_APP_BLOCKNATIVE_KEY,
     FORTMATIC_KEY: process.env.REACT_APP_FORTMATIC_KEY,
     PORTIS_KEY: process.env.REACT_APP_PORTIS_KEY,
-    ZAPPER_API_KEY: process.env.REACT_APP_ZAPPER_API_KEY,
+    ZAPPER_AUTH_TOKEN: encode({ str: `${process.env.REACT_APP_ZAPPER_API_KEY}:`, encoding: 'base64' }),
```
xgambitox, line 21: I would prefer if we leave the env config exactly as it is in `process.env`, just for traceability, and add `ZAPPER_AUTH_TOKEN` in `src/config/constants/index.ts` as we do [here](https://github.com/yearn/yearn-finance-v3/blob/develop/src/config/constants/index.ts#L99)

karelianpie: Agreed, much better as a constant. Refactored with https://github.com/yearn/yearn-finance-v3/pull/651/commits/8d6398acf7b55f38574034bf639eafa482fbcc66

xgambitox (approved):

turtlemoji (approved):

PR #652, opened 2022-05-11
fix: Vault search and filter

huggingbot:
## Description

- Vault search and filter function is not working properly after clicking on the deposit button.
- Steps to reproduce:
  - Go to the vaults page at `/vaults`.
  - Type something in the search box located in the Opportunities box.
  - See vaults get filtered.
  - Click on 'Deposit' on any of the vaults.
  - Dismiss modal.
  - Filter is removed but search box still contains text.
- Fix by running the `useEffect` only when `opportunities` length is changed.
- `setFilteredVaults` will be managed by `SearchInput` and will not be overridden by the `useEffect` call.

## Motivation and Context

- Maintain consistency between search text and filter.

## How Has This Been Tested?

- Tried searching 'usdt', clicking on 'Deposit', then dismissing the modal, and observing the filter behavior.

1) During search
during-search

2) After clicking on 'Deposit' and dismissing modal
after-deposit

Fix will maintain the filter shown in 1) after clicking on deposit and dismissing modal.

### DIFF
--- a/src/client/routes/Vaults/index.tsx
+++ b/src/client/routes/Vaults/index.tsx
@@ -2,7 +2,7 @@ import { useState, useEffect } from 'react';
 import styled from 'styled-components';
 import { useHistory } from 'react-router-dom';
 
-import { useAppSelector, useAppDispatch, useIsMounting, useAppTranslation } from '@hooks';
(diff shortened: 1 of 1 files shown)


xgambitox (approved):
nice one! noticed this issue the other day too

turtlemoji (commented):
approved!

PR #653, opened 2022-05-11
feat: Display only Zapable Tokens by vault WEB-1013

karelianpie:
## Description

Display only Zapable Tokens by vault

## Related Issue

https://linear.app/yearn/issue/WEB-1013/display-only-zapable-tokens-by-vault

## Motivation and Context

Users can identify tokens that can be used to deposit a specific vault.

## How Has This Been Tested?

* Running the FE locally

### DIFF
--- a/src/core/store/modules/tokens/tokens.selectors.ts
+++ b/src/core/store/modules/tokens/tokens.selectors.ts
@@ -41,7 +41,7 @@ const selectSummaryData = createSelector([selectUserTokens], (userTokens) => {
 });
 
 const selectZapInTokens = createSelector([selectUserTokens], (userTokens) => {
-  return userTokens.filter(({ isZapperZapIn }) => isZapperZapIn);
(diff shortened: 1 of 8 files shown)


src/core/store/modules/tokens/tokens.selectors.ts
```diff
@@ -86,26 +86,32 @@ const selectWalletTokensStatus = createSelector(
 
 /* --------------------------------- Helper --------------------------------- */
 interface CreateTokenProps {
-  tokenData: Token;
+  tokenData?: Token;
```
karelianpie, line 89: We were expecting this to be `undefined` below but not handling it when it was, tried to just return `null` in the `createToken()` but it's a bit entangled and was breaking unrelated parts of the code. Potentially worth a refactor in a different PR

karelianpie:
Closing in favour of https://github.com/yearn/yearn-finance-v3/pull/682

PR #660, opened 2022-05-12
feat: toggle for zero apy vault opportunities

### LINKED ISSUE #659: feat: toggle switch for 0 APY
**Elevator Pitch**
add a toggle switch to control when 0% APY vaults are displayed

**Customer Value**
Users, by default, shouldn't see 0 apy vaults

**Definition of Done**

- toggle switch labeled "hide %0 APY vaults"
- displayed in vaults/opportunities for all chains
- on by default (hiding 0 apy vaults)

**Design Approach**
please put switch here
There should be a switch
image

similar to the "dust" of wallets
image

**Additional Context**
Screenshots, UI/UX designs, links or other that will help implementing the feature.

edwinzhng:
## Description

* Adds a toggle to show 0% APY vaults, moved the filter for 0% APY out of the selector and into the page itself
* The toggle position is the default one in the component, which is different from what was requested in the issue but I think this looks fine personally? I can change it if needed

## Related Issue

Fixes #659

## Motivation and Context

* Allows users to show hidden vaults

## How Has This Been Tested?

Manually tested on vaults page

Desktop:
Screen Shot 2022-05-12 at 12 18 22 PM

Screen Shot 2022-05-12 at 12 18 38 PM

Mobile:
Screen Shot 2022-05-12 at 12 27 30 PM

### DIFF
--- a/src/client/routes/Vaults/index.tsx
+++ b/src/client/routes/Vaults/index.tsx
@@ -158,7 +158,7 @@ export const Vaults = () => {
   const dispatch = useAppDispatch();
   const isMounting = useIsMounting();
   // const { isTablet, isMobile, width: DWidth } = useWindowDimensions();
-  const { NETWORK_SETTINGS } = getConfig();
+  const { NETWORK_SETTINGS, CONTRACT_ADDRESSES } = getConfig();
   const walletIsConnected = useAppSelector(WalletSelectors.selectWalletIsConnected);
   const currentNetwork = useAppSelector(NetworkSelectors.selectCurrentNetwork);
   const currentNetworkSettings = NETWORK_SETTINGS[currentNetwork];
@@ -210,6 +210,10 @@ export const Vaults = () => {
       { header: t('dashboard.est-yearly-yield'), Component: <Amount value={estYearlyYeild} input="usdc" /> }
     );
   }
+
+  const filterVaults = (vault: VaultView) => {
+    return toBN(vault.apyMetadata?.net_apy).gt(0) || vault.address === CONTRACT_ADDRESSES.YVYFI;
+  };
 
   return (
     <ViewContainer>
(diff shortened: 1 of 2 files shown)


src/core/store/modules/vaults/vaults.selectors.ts
```diff
@@ -109,9 +109,7 @@ const selectVaultsOpportunities = createSelector([selectLiveVaults], (vaults): V
 
   const depositVaults = vaults.map(({ DEPOSIT, token, ...rest }) => ({ token, ...DEPOSIT, ...rest }));
   const opportunities = depositVaults.filter(
-    (vault) =>
-      toBN(vault.userDeposited).lte(0) &&
-      (toBN(vault.apyMetadata?.net_apy).gt(0) || vault.address === YFI_VAULT_ADDRESS)
+    (vault) => toBN(vault.userDeposited).lte(0) || vault.address === YFI_VAULT_ADDRESS
```
xgambitox, line 112:
```suggestion
    (vault) => toBN(vault.userDeposited).lte(0)
```
this can be removed too

xgambitox:
Position seems fine to me also. Lets send it to prod as it is and we can iterate after if we get better feedback.

src/client/routes/Vaults/index.tsx
```diff
@@ -477,6 +481,8 @@ export const Vaults = () => {
                 />
               }
               searching={opportunities.length > filteredVaults.length}
+              filterLabel="Show 0% APY"
```
turtlemoji, line 484: Can you add the locales in the same place as show dust?

xgambitox: Seems like we never added the "Show dust" to locales

edwinzhng: I can add both

xgambitox: cool, lets include those then

turtlemoji (approved):
if this gets merged we can always add the locales later

xgambitox (approved):

PR #661, opened 2022-05-13
fix: Tx modal token name convention

### LINKED ISSUE #658: fix: tx modal token name convention
**Describe the bug**
The token name on the "from wallet" section should match the token name on the "to vault" section

**To Reproduce**
1. Go to vault details or vault transaction (generally for any curve vault)

**Expected behavior**
The token name convention should match between the wallet and vault deposit

**Screenshots**
image

**Environment Information**

**Additional context**

huggingbot:
## Description

- Updated token name for 'to vault' to use `token.symbol` instead of `displayName`.
- Think using the token symbol would fit better as the display name is already shown in the list.

## Related Issue

- Fixes #658

## Motivation and Context

- Token naming consistency.

## How Has This Been Tested?
- Manually check deposit modal on local environment.

deposit

### DIFF
--- a/src/client/components/app/Transactions/Backscratcher/ClaimTx.tsx
+++ b/src/client/components/app/Transactions/Backscratcher/ClaimTx.tsx
@@ -60,7 +60,7 @@ export const BackscratcherClaimTx: FC<BackscratcherClaimTxProps> = ({ onClose, c
 
   const selectedLabOption = {
     address: selectedLab.address,
-    symbol: selectedTargetToken.name,
+    symbol: selectedTargetToken.symbol,
     icon: selectedTargetToken.icon,
     balance: selectedLab.YIELD.userDeposited,
     balanceUsdc: selectedLab.YIELD.userDepositedUsdc,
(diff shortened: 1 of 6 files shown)


FoxTheSin (dismissed):

xgambitox (commented):
Seems like the same on withdraw (maybe other modals too), worth fixing in same PR if you can take a look.

karelianpie (approved):
LGTM 🙇

xgambitox (approved):

xgambitox:
Seems all good to me. Will merge once @FoxTheSin does some QA and approves too :)

FoxTheSin:
sorry to be nit picky but....
image

FoxTheSin:
this should be how it looks it rolls over
image

FoxTheSin:
(side note it's consistent with mobile version*)

xgambitox:
> sorry to be nit picky but.... image
> 
> steth-weth pool is the only one that didn't get the "..."

If we can fix that it does not cut out from the bottom, that would be the best solution imo, since we will probably have more vaults in the future that might have special characters or spacings that make the text only viewable in two separate lines.

Im fine either by fixing here or creating a separate issue for this one.

karelianpie (approved):

turtlemoji (approved):

PR #674, opened 2022-05-16
feat: service provider controller [WEB-1415]

xgambitox:
## Description

Downgrades simulation/zapper/notify services when providers are not available.

## Related Issue

[Linear Ticket](https://linear.app/yearn/issue/WEB-1415/critical-service-providers-controller)

## Motivation and Context

Users can continue to use the site even on service providers being down, either through a fallback or through a downgraded experience.

## How Has This Been Tested?

Will be tested in preview builds simulating services being not available. (#677 , #678 )


FoxTheSin:
not clear if this is related, doesn't happen on prod (zap with fantom?)
image

FoxTheSin:
also %0 apy are being filtered without toggle switch (is this expected?)

src/core/types/Lab.ts
```diff
@@ -19,6 +19,8 @@ export interface GeneralLabView {
   pricePerShare: string;
   allowZapIn: boolean;
   allowZapOut: boolean;
+  zapInWith?: string;
+  zapOutWith?: string;
```
karelianpie, line 23: Let's narrow these types

```suggestion
  zapInWith?: keyof Token['supported'];
  zapOutWith?: keyof Token['supported'];
```

src/core/types/State.ts
```diff
@@ -33,6 +34,7 @@ export interface RootState {
 
 export interface AppState {
   isInitialized: boolean;
+  servicesEnabled: Record<ServiceId, boolean>;
```
karelianpie, line 37: nit: In terms of extensibility I would say that this could be a bit more than a boolean and also perhaps use the enum naming convention where it's singular instead of plural
```suggestion
  serviceStatus: Record<ServiceId, { enabled: boolean }>;
```
usage
```typescript
const { enabled: notifyEnabled } = app.serviceStatus['notify'];
```
No action needed, just a thought

src/core/types/Token.ts
```diff
@@ -20,6 +20,7 @@ export interface TokenView {
   description: string;
   website: string;
   isZapable: boolean;
+  sourceIsZapper: boolean;
```
karelianpie, line 23: Sounds a bit too specific, could we just have 
```suggestion
  source: SourceType;
```

src/core/types/Yearn-Sdk.ts
```diff
@@ -59,6 +59,8 @@ interface LabMetadata {
   displayName: string;
   displayIcon: string;
   defaultDisplayToken: Address;
+  zapInWith?: string; // TODO: Import zapInType from sdk when updated
+  zapOutWith?: string; // TODO: Import zapOutType from sdk when updated
```
karelianpie, line 63: Ahh now I see this TODO, we could also just use the same type I've set in there for now just in case 
```typescript
keyof Token["supported"]
```

src/core/types/Vault.ts
```diff
@@ -35,6 +35,8 @@ export interface GeneralVaultView {
   pricePerShare: string;
   allowZapIn: boolean;
   allowZapOut: boolean;
+  zapInWith?: string;
+  zapOutWith?: string;
```
karelianpie, line 39: Should we add a TODO here as well? Maybe worth to add the same type I've mentioned here https://github.com/yearn/yearn-finance-v3/pull/674/files#r875332401 for now just in case

src/core/types/General.ts
```diff
@@ -8,4 +8,6 @@ export type DataType = 'amount' | 'percent' | 'usd';
 
 export type Route = 'portfolio' | 'vaults' | 'vault' | 'labs';
 
+export type ServiceId = 'zapper' | 'tenderly' | 'notify';
```
karelianpie, line 11: I think we should be a bit more explicit on what type of services these are
```suggestion
export type ExternalServiceId = 'zapper' | 'tenderly' | 'notify';
```

src/core/store/modules/tokens/tokens.selectors.ts
```diff
@@ -106,6 +87,7 @@ export function createToken(props: CreateTokenProps): TokenView {
     description: tokenData?.metadata?.description ?? '',
     website: tokenData?.metadata?.website ?? '',
     isZapable: tokenData?.supported.zapper ?? false,
```
karelianpie, line 89: Should we add other supported zaps here?
```suggestion
    isZapable: tokenData?.supported.zapper || tokenData?.supported.ftmApeZap ?? false,
```

src/core/store/modules/tokens/tokens.selectors.ts
```diff
@@ -106,6 +87,7 @@ export function createToken(props: CreateTokenProps): TokenView {
...
+    sourceIsZapper: tokenData?.dataSource === 'zapper',
```
karelianpie, line 90: As mentioned in https://github.com/yearn/yearn-finance-v3/pull/674#discussion_r875331720 could we just have the source here? When we want to extend this to other data sources it would be better to have a more extendable data structure

src/core/store/modules/tokens/tokens.selectors.ts
```diff
@@ -14,21 +17,22 @@ const selectUserTokensStatusMap = (state: RootState) => state.tokens.statusMap;
...
+    const { userTokensAddresses, userTokensMap, userTokensAllowancesMap } = user;
+    const tokens = userTokensAddresses
+      .filter((address) => !!tokensMap[address])
+      .map((address) => {
+        const tokenData = tokensMap[address];
+        const userTokenData = userTokensMap[address];
+        const allowancesMap = userTokensAllowancesMap[address] ?? {};
+        return createToken({ tokenData, userTokenData, allowancesMap });
+      });
+    return tokens.filter((token) => toBN(token.balance).gt(0) && (!token.sourceIsZapper || servicesEnabled.zapper));
```
karelianpie, line 33: Are we tying this too much to zapper?

src/core/store/modules/app/app.actions.ts
```diff
@@ -117,6 +125,47 @@ const getUserAppData = createAsyncThunk<void, { network: Network; route: Route;
   }
 );
 
+/* -------------------------------------------------------------------------- */
+/*                                  Services                                  */
+/* -------------------------------------------------------------------------- */
+
+const checkServicesStatus = createAsyncThunk<void, void, ThunkAPI>(
```
karelianpie, line 132:
```suggestion
const checkExternalServicesStatus = createAsyncThunk<void, void, ThunkAPI>(
```

src/core/store/modules/app/app.actions.ts
```diff
@@ -117,6 +125,47 @@ const getUserAppData = createAsyncThunk<void, { network: Network; route: Route;
...
+  'app/checkServicesStatus',
+  async (_arg, { dispatch, extra }) => {
+    const { YEARN_ALERTS_API } = extra.config;
+    try {
+      const servicesStatusResponse = await get(`${YEARN_ALERTS_API}/health`);
+      if (servicesStatusResponse.status !== 200) throw new Error('Service status provider not currently accesible');
```
karelianpie, line 138: typo, it's `accessible`

src/core/store/modules/app/app.actions.ts
```diff
@@ -117,6 +125,47 @@ const getUserAppData = createAsyncThunk<void, { network: Network; route: Route;
...
+/* -------------------------------------------------------------------------- */
+/*                                  Services                                  */
+/* -------------------------------------------------------------------------- */
+
+const checkServicesStatus = createAsyncThunk<void, void, ThunkAPI>(
+  'app/checkServicesStatus',
+  async (_arg, { dispatch, extra }) => {
+    const { YEARN_ALERTS_API } = extra.config;
+    try {
+      const servicesStatusResponse = await get(`${YEARN_ALERTS_API}/health`);
```
karelianpie, line 137: nit: A bit cleaner IMO to destructure here
```suggestion
      const { status, data } = await get(`${YEARN_ALERTS_API}/health`);
```

src/core/store/modules/app/app.actions.ts
```diff
@@ -117,6 +125,47 @@ const getUserAppData = createAsyncThunk<void, { network: Network; route: Route;
...
+      if (servicesStatusResponse.status !== 200) throw new Error('Service status provider not currently accesible');
+
+      const errorMessageTemplate =
+        'service is currently experiencing technical issues and have been temporarily disabled. We are sorry for the inconveniences, we are working towards issues been resolved soon.';
```
karelianpie, line 141: The latter part of the sentence can be improved, maybe
```suggestion
        'service is currently experiencing technical issues and have been temporarily disabled. We apologize for any inconvenience this may have caused and our team is actively working on resolving these issues.';
```

src/core/store/modules/app/app.actions.ts
```diff
@@ -117,6 +125,47 @@ const getUserAppData = createAsyncThunk<void, { network: Network; route: Route;
...
+      const downgradedServicesMessages = [];
+      const { zapper, simulations } = servicesStatusResponse.data;
```
karelianpie, line 143: Are we not getting the `notify` service here?

src/core/store/modules/alerts/alerts.actions.ts
```diff
@@ -11,15 +12,16 @@ export interface AlertProps {
...
-const closeAlert = createAction<{ alertId: number }>('alerts/closeAlert');
+const closeAlert = createAction<{ alertId: string }>('alerts/closeAlert');
 
 const openAlert = createAsyncThunk<
   { alert: Alert },
   { message: string; type?: AlertTypes; persistent?: boolean; timeout?: number },
   ThunkAPI
->('alerts/openAlert', async ({ message, type, persistent, timeout }, { getState, dispatch }) => {
+>('alerts/openAlert', async ({ message, type, persistent, timeout }, { dispatch }) => {
+  const id = getRandomId();
```
karelianpie, line 22: 👍

src/core/services/TransactionService.ts
```diff
@@ -101,27 +101,36 @@ export class TransactionServiceImpl implements TransactionService {
   public handleTransaction = async ({
     network,
     tx,
+    useExternalService = false,
```
karelianpie, line 104: I don't think this is needed, it's already undefined by default
```suggestion
    useExternalService,
```

src/core/services/TransactionService.ts
```diff
@@ -101,27 +101,36 @@ export class TransactionServiceImpl implements TransactionService {
...
-        } else {
-          const { update, dismiss } = notify.notification({
-            eventCode: 'txSentCustom',
-            type: 'pending',
-            message: 'Your transaction has been sent to the network',
-          });
-          updateNotification = update;
-          dismissNotification = dismiss;
-        }
+    let notifyServiceFailed = false;
```
karelianpie, line 113: Same here, this will be `undefined` and thus falsy
```suggestion
    let notifyServiceFailed;
```

src/core/services/TransactionService.ts
```diff
@@ -101,27 +101,36 @@ export class TransactionServiceImpl implements TransactionService {
...
+    if (renderNotification && useNotifyService) {
+      try {
+        notify.hash(tx.hash);
+      } catch (error: any) {
+        notifyServiceFailed = true;
+        console.error(error);
       }
+    }
+
+    if (renderNotification && (!useNotifyService || notifyServiceFailed)) {
```
karelianpie, line 123: I'm not sure I follow this, so we are checking `!useNotifyService || notifyServiceFailed` and then using `notify.notification()`? Isn't that using the notify service?

src/client/components/app/Transactions/DepositTx.tsx
```diff
@@ -54,6 +55,9 @@ export const DepositTx: FC<DepositTxProps> = ({
   const [debouncedAmount, isDebouncePending] = useDebounce(amount, 500);
   const [txCompleted, setTxCompleted] = useState(false);
   const [isFetchingAllowance, setIsFetchingAllowance] = useState(false);
+  const servicesEnabled = useAppSelector(AppSelectors.selectServicesEnabled);
+  const simulationsEnabled = servicesEnabled.tenderly;
```
karelianpie, line 59: nit: Wondering if we should be more explicit on which service we are using so that we know that it's tenderly that is down looking at the code
```suggestion
  const tenderlyEnabled = servicesEnabled.tenderly;
```

src/client/components/app/Transactions/DepositTx.tsx
```diff
@@ -137,14 +150,16 @@ export const DepositTx: FC<DepositTxProps> = ({
...
 
-    dispatch(
-      VaultsActions.getExpectedTransactionOutcome({
-        transactionType: 'DEPOSIT',
-        sourceTokenAddress: selectedSellTokenAddress,
-        sourceTokenAmount: toWei(debouncedAmount, selectedSellToken.decimals),
-        targetTokenAddress: selectedVault.address,
-      })
-    );
+    if (simulationsEnabled) {
```
karelianpie, line 153: nitY I always think it's cleaner to just early return than checking where the curly brace ends

```suggestion
    if (!simulationsEnabled) {
        return;
    }
```

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
```diff
@@ -48,6 +49,9 @@ export const LabDepositTx: FC<LabDepositTxProps> = ({ onClose }) => {
   const [debouncedAmount, isDebouncePending] = useDebounce(amount, 500);
   const [txCompleted, setTxCompleted] = useState(false);
   const [isFetchingAllowance, setIsFetchingAllowance] = useState(false);
+  const servicesEnabled = useAppSelector(AppSelectors.selectServicesEnabled);
+  const simulationsEnabled = servicesEnabled.tenderly;
+  const zapperEnabled = servicesEnabled.zapper;
```
karelianpie, line 54:
```suggestion
  const { tenderly: simulationsEnabled, zapper: zapperEnabled } = servicesEnabled;
```

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
```diff
@@ -107,14 +120,17 @@ export const LabDepositTx: FC<LabDepositTxProps> = ({ onClose }) => {
...
 
-    dispatch(
-      VaultsActions.getExpectedTransactionOutcome({
-        transactionType: 'DEPOSIT',
-        sourceTokenAddress: selectedSellTokenAddress,
-        sourceTokenAmount: toWei(debouncedAmount, selectedSellToken.decimals),
-        targetTokenAddress: selectedLab.address,
-      })
-    );
+    if (simulationsEnabled) {
```
karelianpie, line 123:
```suggestion
    if (!simulationsEnabled) {
        return;
    }
```

src/client/components/app/Transactions/Labs/LabWithdrawTx.tsx
```diff
@@ -40,21 +42,20 @@ export const LabWithdrawTx: FC<LabWithdrawTxProps> = ({ onClose, children, ...pr
   const [amount, setAmount] = useState('');
   const [debouncedAmount, isDebouncePending] = useDebounce(amount, 500);
   const [txCompleted, setTxCompleted] = useState(false);
+  const servicesEnabled = useAppSelector(AppSelectors.selectServicesEnabled);
+  const simulationsEnabled = servicesEnabled.tenderly;
+  const zapperEnabled = servicesEnabled.zapper;
```
karelianpie, line 47:
```suggestion
  const { tenderly: simulationsEnabled, zapper: zapperEnabled } = servicesEnabled;
```

src/client/components/app/Transactions/Labs/LabWithdrawTx.tsx
```diff
@@ -96,7 +107,7 @@ export const LabWithdrawTx: FC<LabWithdrawTxProps> = ({ onClose, children, ...pr
 
   useEffect(() => {
     if (!selectedLab || !selectedTargetTokenAddress) return;
-    if (toBN(debouncedAmount).gt(0) && !inputError) {
+    if (simulationsEnabled && toBN(debouncedAmount).gt(0) && !inputError) {
```
karelianpie, line 110: A lot going on here

src/client/components/app/Transactions/WithdrawTx.tsx
```diff
@@ -42,21 +43,19 @@ export const WithdrawTx: FC<WithdrawTxProps> = ({ header, onClose, children, ...
   const [amount, setAmount] = useState('');
   const [debouncedAmount, isDebouncePending] = useDebounce(amount, 500);
   const [txCompleted, setTxCompleted] = useState(false);
+  const servicesEnabled = useAppSelector(AppSelectors.selectServicesEnabled);
+  const simulationsEnabled = servicesEnabled.tenderly;
+  const zapperEnabled = servicesEnabled.zapper;
```
karelianpie, line 48:
```suggestion
  const { tenderly: simulationsEnabled, zapper: zapperEnabled } = servicesEnabled;
```

src/core/types/Token.ts
xgambitox: Did not wanted to propagate the typing across the app, since `SourceType` comes from SDK, and thought it was best to convert it to a primitive for the specific use case we needed, since no other source is needed till now. Maybe its worth to change it if we do have a use case in the future for the source... wdyt?

src/core/store/common/tokensByAsset.selectors.ts
```diff
@@ -0,0 +1,86 @@
...
+      const { userTokensAddresses, userTokensMap, userTokensAllowancesMap } = tokensUser;
+      const assetData = vaultsMap[assetAddress] ? vaultsMap[assetAddress] : labsMap[assetAddress];
+      if (!assetData) return [];
+
+      const zapperDisabled = !servicesEnabled.zapper && assetData.metadata.zapInWith === 'zapperZapIn';
+      const depositTokenAddresses: Address[] = [];
+      let mainVaultToken: Address;
+      if (zapperDisabled) {
+        mainVaultToken = assetData.token;
+        depositTokenAddresses.push(mainVaultToken);
```
karelianpie, line 34: This is being done in both if/else blocks, can we move it out

src/core/store/common/tokensByAsset.selectors.ts
```diff
@@ -0,0 +1,86 @@
...
+          const tokenData = tokensMap[address];
+          const userTokenData = userTokensMap[address];
+          const allowancesMap = userTokensAllowancesMap[address] ?? {};
+          return createToken({ tokenData, userTokenData, allowancesMap });
+        });
+      return tokens.filter((token) => toBN(token.balance).gt(0) || token.address === mainVaultToken);
+    })
+);
+
+export const selectWithdrawTokenOptionsByAsset = createSelector(
```
karelianpie, line 53: Both of these functions seem to have quite a bit in common, could we extract some of the logic to a common function so that we can then refactor more easily?

Screen Shot 2022-05-18 at 9 54 58 am

src/core/store/common/tokensByAsset.selectors.ts
```diff
@@ -0,0 +1,86 @@
...
+  [selectVaultsMap, selectLabsMap, selectTokensMap, selectTokensUser, selectServicesEnabled],
+  (vaultsMap, labsMap, tokensMap, tokensUser, servicesEnabled) =>
+    memoize((assetAddress?: string): TokenView[] => {
+      if (!assetAddress) return [];
+
+      const { ZAP_OUT_TOKENS } = getConfig();
```
karelianpie, line 59: This is only used in line 73, let's move it inside the `else` block

karelianpie (changes requested):
Overall makes sense to me, left some comments

src/core/types/Token.ts
karelianpie: Yeah, that's what I was thinking, in the future if more sources come we'd have to add props `sourceIsX`, `sourceIsY`, etc. If we don't foresee that happening then we can go with this, but we're reducing extensibility.

src/core/types/Lab.ts
xgambitox: Added here https://github.com/yearn/yearn-finance-v3/pull/674/commits/7170e91d3685d21e399b3e874c82d785876763dd. Had a TODO after about this, but better change it now and refactor later with the import.

src/core/types/Token.ts
xgambitox: Yeah, agree. If more sources are needed later, then it would be better to refactor this part later as you pointed rather than adding 'sourceIsX'.

src/core/types/Yearn-Sdk.ts
xgambitox: refactored here https://github.com/yearn/yearn-finance-v3/pull/674/commits/b70f2e7d0cd1238378dc4a49fa0fdee00c0bd66b

src/core/types/Vault.ts
xgambitox: yep, changed here https://github.com/yearn/yearn-finance-v3/pull/674/commits/b70f2e7d0cd1238378dc4a49fa0fdee00c0bd66b and here https://github.com/yearn/yearn-finance-v3/commit/7170e91d3685d21e399b3e874c82d785876763dd as suggested.

src/core/types/General.ts
xgambitox: Agree, changed to better naming here https://github.com/yearn/yearn-finance-v3/pull/674/commits/61331273545f2bb6eb771a5df386c3825a40c640

src/core/store/modules/tokens/tokens.selectors.ts
xgambitox: imo, this should be meant for the PR that implements custom zaps 😁 

Also, `supported.zapper` can start to get deprecated in favor of the explicit zapper contract, like `supported.zapperZapIn`

src/core/store/modules/tokens/tokens.selectors.ts
xgambitox: what I dont like about extending external types like `dataSource`, is that we couple our app to it, and if it gets changed for some reason (even though its our own SDK), it breaks all layers of the app. Here, we only need to change the adapter layer, the selector in this case.

src/core/store/modules/tokens/tokens.selectors.ts
xgambitox: In this case we are only trying to switch zapper on/off, so its intended to be for zapper and not all zaps.

I do think this is not the best place for it, and maybe a common selector would be better, since it mixes two slices of state now... but did not want to overcomplicate the changes.

src/core/store/modules/app/app.actions.ts
xgambitox: changed here https://github.com/yearn/yearn-finance-v3/pull/674/commits/3e0467ae877f70c98224e1f1f1a5ff048d2faea6

src/core/store/modules/app/app.actions.ts
xgambitox: fixed https://github.com/yearn/yearn-finance-v3/pull/674/commits/ac85e1e489ab9ec7bbd482f5d227a5a682bdcd01 😁

src/core/store/modules/app/app.actions.ts
xgambitox: refactored here https://github.com/yearn/yearn-finance-v3/pull/674/commits/06ec326e7ff0dd9c530a2f337abbf14fc3ccf217

src/core/store/modules/app/app.actions.ts
xgambitox: changed it here https://github.com/yearn/yearn-finance-v3/pull/674/commits/af9874b883993b37d88d3df4cd351647b1a091e3

close enough?

src/core/store/modules/app/app.actions.ts
xgambitox: Could not find any way to consume that Notify service is not working, unless the service itself fails for the user or we manually turn it off via PR unfortunately :S

src/core/services/TransactionService.ts
xgambitox: much nit?

changed here https://github.com/yearn/yearn-finance-v3/pull/674/commits/cc4c6b9e0a0ee50a21a9454875983308ccc16afc XD

src/core/services/TransactionService.ts
xgambitox: much nit?

changed here https://github.com/yearn/yearn-finance-v3/commit/cc4c6b9e0a0ee50a21a9454875983308ccc16afc XD

src/core/services/TransactionService.ts
xgambitox: `notify.notification()` is meant for custom messages, more as a UI component, and should not be utilizing Blocknative service AFAIK. Remember reading that somewhere but cant find it now https://docs.blocknative.com/notify#notification.

We actually do this for fantom, since its disabled by default https://github.com/yearn/yearn-finance-v3/blob/develop/src/config/constants/index.ts#L67

src/client/components/app/Transactions/DepositTx.tsx
xgambitox: did thought about it, but also thought we should have a fallback service in the future like

```
const simulationsEnabled = servicesEnabled.tenderly || servicesEnabled.alternateService;
```

And only turn off simulations when both are down...

src/client/components/app/Transactions/DepositTx.tsx
xgambitox: It can be done if I also change the order of the action dispatched after this statement. Thought it was clearer just to place it as it is. Guess it is matter of preference on this one.

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
xgambitox: same as above about order of the next statement

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
xgambitox: placed them in different line just because of what mentioned above about
```
const simulationsEnabled = servicesEnabled.tenderly || servicesEnabled.alternateService;
```

src/client/components/app/Transactions/Labs/LabWithdrawTx.tsx
xgambitox: same as above

src/core/store/common/tokensByAsset.selectors.ts
xgambitox: the thing is order matter on this one since we want to have the `mainVaultToken` as first element in the array.

src/core/store/modules/tokens/tokens.selectors.ts
karelianpie: Yes, let's leave it for the other PR

src/core/store/common/tokensByAsset.selectors.ts
xgambitox: maybe the last part of both where `tokens` are mapped, but not sure its worth it.

src/core/store/modules/app/app.actions.ts
karelianpie: Sounds better 👍

src/core/services/TransactionService.ts
karelianpie: It wasn't a nit :) I think it adds a bit of confusing, implying that `useExternalService` is required when it's not

src/core/store/common/tokensByAsset.selectors.ts
xgambitox: moved here https://github.com/yearn/yearn-finance-v3/pull/674/commits/795eb535cedbb008a925adb6641da8b09fa09865

src/core/services/TransactionService.ts
karelianpie: Ah I see, so what we're checking with `notifyServiceFailed` is if `blockNativeNotifyServiceFailed` -- if so, I think it's worth to rename as it's a bit confusing. Initially I was wondering why we are using the "notify" service by calling `notification()` on it if it failed for example

src/client/components/app/Transactions/DepositTx.tsx
karelianpie: In that case, I'd try and centralize that knowledge and have an ExternalServiceStatus service that understands the domain of external services, and then we would have a function on there `isSimulationEnabled()` that would check `tenderly` and then the `alternateService`. It feels a bit too much for the `DepositTx` component to have such knowledge IMO

src/client/components/app/Transactions/DepositTx.tsx
karelianpie: Happy to leave it as is

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
karelianpie: I've left my opinion in the other conversation, but if we go with that for simulations, then shouldn't we do the same for zaps? As in, have a `const zapsEnabled = servicesEnabled.zapper || servicesEnabled.alternativeZap`

I still think this should live in a centralized place and not scattered around, `LabDepositTx` knows too much about what is an enabled simulation etc.

src/core/services/TransactionService.ts
xgambitox: yeah, probably makes sence to refactor to better naming. will do

src/core/store/common/tokensByAsset.selectors.ts
karelianpie: It's not a problem, you can use `unshift` or just use the spread operator to get it done :)

src/core/store/common/tokensByAsset.selectors.ts
karelianpie: Let's leave it, usually there's a rule of 3... when it's repeated 3x it's the time to refactor.

src/core/services/TransactionService.ts
xgambitox: changed here https://github.com/yearn/yearn-finance-v3/pull/674/commits/e4da2120c74fded35f559f4fe787f85e84f99719

karelianpie: I was also thinking we should refactor the `notifyServiceFailed` to `blockNativeNotifyServiceFailed`

src/core/store/common/tokensByAsset.selectors.ts
xgambitox: done here https://github.com/yearn/yearn-finance-v3/pull/674/commits/56d21c06d4d0beb30c4cc0cff93519e73646b743 :)

src/client/components/app/Transactions/DepositTx.tsx
xgambitox: yeah, you are right. I actually though of adding that since we will need it. There should be a common selector that aggregates all those, specially since there will also be another state for user settings, where user can disable external services manually if they wish, for privacy reasons. It just felt out of scope right now.

Think my example was wrong then, and meant to be something like:
```ts
const simulationsEnabled = useAppSelector(selectIsSimulationEnabled);
```
And that is why I chose that specific naming

src/client/components/app/Transactions/Labs/LabDepositTx.tsx
xgambitox: yeah, expanded on your comment above

src/client/components/app/Transactions/DepositTx.tsx
karelianpie: Sounds good, let's leave it for now

src/core/services/TransactionService.ts
xgambitox: noticed it too and edited my comment after :P

src/core/store/common/tokensByAsset.selectors.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      const zapperDisabled = !servicesEnabled.zapper && assetData.metadata.zapInWith === 'zapperZapIn';
+      const depositTokenAddresses: Address[] = [];
+      let mainVaultToken: Address;
+      if (zapperDisabled) {
+        mainVaultToken = assetData.token;
+      } else {
+        mainVaultToken = assetData.metadata.defaultDisplayToken;
+        depositTokenAddresses.push(...userTokensAddresses.filter((address) => address !== mainVaultToken));
+      }
```
karelianpie, line 37: Now that we've simplified the logic inside the `if/else` we can remove it all together, and also the `let` and make it a `const` for readability, something like:

```suggestion
      const mainVaultToken: Address = zapperDisabled ? assetData.token : assetData.metadata.defaultDisplayToken;

      if (!zapperDisabled) {
        const { ZAP_OUT_TOKENS } = getConfig();
        withdrawTokenAddresses.push(...ZAP_OUT_TOKENS.filter((address) => address !== mainVaultToken));
      }
```

xgambitox: great suggestion. refactored here https://github.com/yearn/yearn-finance-v3/pull/674/commits/0e496c73bdac14363d64c3907f39c3108889835b

karelianpie: Ah, nice idea with the refactor 👍  much better!

karelianpie (approved):
LGTM

FoxTheSin (approved):

PR #684, opened 2022-05-18
feat: use private zapper api key

xgambitox:
## Description

Initiate sdk instance with zapper private api key

## Motivation and Context

Zapper will be deprecating soon the public API key that is used by default in the SDK.

## How Has This Been Tested?

To be tested in preview build

### DIFF
--- a/src/core/frameworks/yearnSdk/index.ts
+++ b/src/core/frameworks/yearnSdk/index.ts
@@ -7,7 +7,7 @@ export class YearnSdkImpl implements YearnSdk {
   private instances: Map<Network, Yearn<SdkNetwork>> = new Map<Network, Yearn<SdkNetwork>>();
 
   constructor({ web3Provider, config }: { web3Provider: Web3Provider; config: Config }) {
-    const { SUPPORTED_NETWORKS, CONTRACT_ADDRESSES, YEARN_SUBGRAPH_ID, YEARN_SUBGRAPH_KEY } = config;
+    const { SUPPORTED_NETWORKS, CONTRACT_ADDRESSES, YEARN_SUBGRAPH_ID, YEARN_SUBGRAPH_KEY, ZAPPER_API_KEY } = config;
 
     const isLedger = isLedgerLive();
     SUPPORTED_NETWORKS.forEach((network) => {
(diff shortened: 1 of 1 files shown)


karelianpie:
> Zapper will be deprecating soon the public API key 

They did not mention this, they said they would give plenty of time for people to move to the private key instead. Or did something change?

xgambitox:
> > Zapper will be deprecating soon the public API key
> 
> They did not mention this, they said they would give plenty of time for people to move to the private key instead. Or did something change?

No need to not doing it now. Here what is stated on the [article](https://medium.com/zapper-protocol/updates-to-zappers-public-apis-4933a2967f57):

"If you haven’t already done so, we highly recommend you [request a private API key](https://zapperfi.zendesk.com/hc/en-us/requests/new), as the public key will continue to be heavily throttled."

"we are planning to deprecate the public API key entirely, and ask the community to leverage private keys for querying data. We do not have a date set for the sunsetting of the public key, and will give folks plenty of time to migrate."

src/core/frameworks/yearnSdk/index.ts
```diff
@@ -7,7 +7,7 @@ export class YearnSdkImpl implements YearnSdk {
   private instances: Map<Network, Yearn<SdkNetwork>> = new Map<Network, Yearn<SdkNetwork>>();
 
   constructor({ web3Provider, config }: { web3Provider: Web3Provider; config: Config }) {
-    const { SUPPORTED_NETWORKS, CONTRACT_ADDRESSES, YEARN_SUBGRAPH_ID, YEARN_SUBGRAPH_KEY } = config;
+    const { SUPPORTED_NETWORKS, CONTRACT_ADDRESSES, YEARN_SUBGRAPH_ID, YEARN_SUBGRAPH_KEY, ZAPPER_API_KEY } = config;
```
FiboApe, line 10: its not very private if we have it on our code right? :P

karelianpie:
> and will give folks plenty of time to migrate

this was what I remember, but I guess good to update earlier.

karelianpie (approved):

FoxTheSin:
zapping from link to dai looked good, I rejected transaction in wallet.

FoxTheSin (approved):

Issue #683, opened 2022-05-18
fix: susd


FoxTheSin:
**Describe the bug**
when sUSD in wallet (and potentially other tokens?) the deposit modal doesn't appear when clicking deposit button.

**To Reproduce**
1. Go to portfolio with susd
2. click on deposit next to susd token

**Expected behavior**
The deposit modal should appear and token should be zappable into vault (or, if unzapable, an error message should pop up in modal that states that token can't be zap'd)

**Screenshots**
normal
![image](https://user-images.githubusercontent.com/79029523/169094513-cc7f89d6-2a1e-4f20-a497-1112961874a9.png)

susd
![image](https://user-images.githubusercontent.com/79029523/169094700-8006caec-6164-4b7c-a1fd-5f5eaa2000d5.png)

**Environment Information**

**Additional context**
susd contract: 0x57Ab1ec28D129707052df4dF418D58a2D46d5f51

xgambitox:
Yeah. First step is to determine if sUSD is supported by zapper for zapping in into vaults. If its not supported, then we should not be displaying it at all in the portfolio view, since we want to display only actionable tokens within yearn app.

xgambitox:
Maybe there was an sUSD vault that is now deprecated, and that is why sUSD is treated as a valid underlying token for direct deposits?

PR #688, opened 2022-05-19
fix: Filter zapable token for wallet

huggingbot:
## Description

- Added `filter` to filter only zapable tokens
- Checked that sUSD is not zapable in yearn-sdk
- Checked that sUSD yVault is not deprecated

## Related Issue

- #683

## Motivation and Context

- Improve user experience by not showing unzapable vaults

## How Has This Been Tested?

- Locally by confirming [sUSD ](https://etherscan.io/address/0x57ab1ec28d129707052df4df418d58a2d46d5f51)not appearing in the wallet container

- [sUSD yVault](https://etherscan.io/address/0xa5cA62D95D24A4a350983D5B8ac4EB8638887396#code) is shown to be not deprecated by checking [AddressesGenerator_VAULT_V2](https://etherscan.io/address/0x437758d475f70249e03eda6be23684ad1fc375f0#readContract)'s `assetDeprecated` and confirming vault is being fetched by checking `assetsAddresses`

AddressesGenerator_VAULT_V2_0x437758d475f70249e03eda6be23684ad1fc375f0

- sUSD token not showing anymore

susd

### DIFF
--- a/src/client/routes/Portfolio/index.tsx
+++ b/src/client/routes/Portfolio/index.tsx
@@ -264,13 +264,15 @@ export const Portfolio = () => {
               grow: '1',
             },
           ]}
-          data={userTokens.map((token) => ({
-            ...token,
-            displayName: token.symbol,
-            displayIcon: token.icon ?? '',
-            tokenBalance: normalizeAmount(token.balance, token.decimals),
-            invest: null,
-          }))}
+          data={userTokens
+            .filter((token) => token.isZapable)
+            .map((token) => ({
+              ...token,
+              displayName: token.symbol,
+              displayIcon: token.icon ?? '',
+              tokenBalance: normalizeAmount(token.balance, token.decimals),
+              invest: null,
+            }))}
           initialSortBy="balanceUsdc"
           wrap
           filterBy={filterDustTokens}


turtlemoji (approved):

FiboApe (approved):

src/client/routes/Portfolio/index.tsx
```diff
@@ -264,13 +264,15 @@ export const Portfolio = () => {
...
           ]}
-          data={userTokens.map((token) => ({
-            ...token,
-            displayName: token.symbol,
-            displayIcon: token.icon ?? '',
-            tokenBalance: normalizeAmount(token.balance, token.decimals),
-            invest: null,
-          }))}
+          data={userTokens
+            .filter((token) => token.isZapable)
```
karelianpie, line 268: I think we will deprecate this property very soon

karelianpie (changes requested):
We will be using the new zap properties, which will most likely deprecate the `isZapable` property.

karelianpie:
> > We will be using the new zap properties, which will most likely deprecate the `isZapable` property.
> 
> I assume the new prop for the `TokenView` type will be `sourceIsZapper`. Could it just replace `isZapable`? Because I realize there are tokens where their `isZapable` is `true`, but their `sourceIsZapper` is `false`. Thanks

No, `sourceIsZapper` will probably be deprecated as well. We're changing the whole way this is working, it's currently in progress as I'm working on the frontend bit as we speak, we'll most likely use the `supported` prop which in turn has `zapperZapIn`/`zapperZapOut`/`ftmApeZap` and other zaps. Thanks for having a look at this @huggingbot but let's have this on hold for the time being.

PR #691, opened 2022-05-21
feat: Display only Zappable Tokens by vault WEB-1013

karelianpie:
## Description

Display only Zapable Tokens by vault

## Related Issue

https://linear.app/yearn/issue/WEB-1013/display-only-zapable-tokens-by-vault

## Motivation and Context

Users can identify tokens that can be used to deposit a specific vault.

## How Has This Been Tested?

* Running the FE locally


src/core/types/Token.ts
```diff
@@ -22,9 +21,9 @@ export interface TokenView {
...
-}
-
-export interface TokenDynamicData {
-  address: Address;
-  priceUsdc: string;
+  supported: {
+    zapperZapIn?: boolean;
+    zapperZapOut?: boolean;
+    ftmApeZap?: boolean;
+  };
```
xgambitox, line 28: We should look into exporting from SDK the `zapContractId` so that we dont need to touch this file when new zaps get added in SDK. Maybe just add a TODO for now and we can create a ticket for SDK.

xgambitox: Saw the TODO directly in the code where this is needed, so no need to place here again :)

src/core/store/common/tokensByAsset.selectors.ts
```diff
@@ -41,7 +49,11 @@ export const selectDepositTokenOptionsByAsset = createSelector(
           const allowancesMap = userTokensAllowancesMap[address] ?? {};
           return createToken({ tokenData, userTokenData, allowancesMap });
         });
-      return tokens.filter((token) => toBN(token.balance).gt(0) || token.address === mainVaultToken);
+      return tokens.filter(
+        (token) =>
+          isSupportedToken({ assetData, token, mainVaultToken, zapperDisabled, zapType: 'zapInWith' }) &&
+          toBN(token.balance).gt(0)
+      );
```
xgambitox, line 56: Seems this part is not correct since `mainVaultToken` should be returned even if it does not have balance.

How about something like:
```suggestion
      return tokens.filter(
        (token) => {
           const isZappable = token.supported[assetData.metadata.zapInWith] && toBN(token.balance).gt(0);
           return isZappable || token.address === mainVaultToken;
        }
      );
```

src/core/store/common/tokensByAsset.selectors.ts
```diff
@@ -71,6 +83,19 @@ export const selectWithdrawTokenOptionsByAsset = createSelector(
           const allowancesMap = userTokensAllowancesMap[address] ?? {};
           return createToken({ tokenData, userTokenData, allowancesMap });
         });
-      return tokens;
+      return tokens.filter((token) =>
+        isSupportedToken({ assetData, token, mainVaultToken, zapperDisabled, zapType: 'zapOutWith' })
+      );
```
xgambitox, line 88: How about something like:
```suggestion
      return tokens.filter((token) => {
         const isZappable = token.supported[assetData.metadata.zapOutWith];
         return isZappable || token.address === mainVaultToken;
      }
      );
```

src/core/store/common/tokensByAsset.selectors.ts
karelianpie: I think that works, I've refactored to that other function because at the minute we need to cast and it becomes noisy. What do you think of https://github.com/yearn/yearn-finance-v3/pull/691/commits/58ab9e2b15f782a16eb0ccee76556690f4f62323

src/core/store/common/tokensByAsset.selectors.ts
karelianpie: Refactored here https://github.com/yearn/yearn-finance-v3/pull/691/commits/58ab9e2b15f782a16eb0ccee76556690f4f62323 so we can have all the logic in one place

xgambitox:
> Since this PR should enable FTM token to appear as an option for deposits in WFTM vault on fantom, we should only merge once we have enabled that functionality on SDK.

Since recently we temporarily disabled zaps in other networks via https://github.com/yearn/yearn-finance-v3/pull/697 , should be good to merge now.

Tested everything and seems to work as expected now :)

It effectively now removes Curve YFI-ETH, Curve EURN and Curve USDPax from being zappable.

xgambitox (approved):

PR #694, opened 2022-05-22
feat: zappable tokens shortcut [WEB-1524]

turtlemoji:
## Description
Should add, in vault details and modal, a quick action to change between tokens with available amounts

image

## How Has This Been Tested?
Local env, desktop, mobile down to 375px.

## Screenshots (if appropriate):
Result:
image

Mobile:
image

PR #695, opened 2022-05-22
fix: Modal zap layout on iphone se

### LINKED ISSUE #692: Modal zap breaking on Iphone SE
Breaking in Iphone SE
image

_Originally posted by @turtlemoji in https://github.com/yearn/yearn-finance-v3/issues/690#issuecomment-1133523318_

huggingbot:
## Description

- Added CSS changes for devices with `400px` width or lower
- `400px` is used as that's the point where the text would wrap to the second line
- Removed padding so that font size can be at `1.3rem`, else with the existing padding, font size has to be reduced to `1.2rem`
- Centered the text so that the container's left and right are aligned
- `341px` will be the next point where the text would wrap again

## Related Issue

- Fixes #692

## Motivation and Context

- Better display for iPhone SE users

## How Has This Been Tested?

- Locally by viewing the updated modal layout

zap

### DIFF
--- a/src/client/components/app/Transactions/components/TxTokenInput.tsx
+++ b/src/client/components/app/Transactions/components/TxTokenInput.tsx
@@ -130,6 +130,13 @@ const ZapMessageContainer = styled.div`
   padding: ${({ theme }) => theme.layoutPadding};
   font-size: 1.4rem;
   width: 100%;
+
+  @media (max-width: 400px) {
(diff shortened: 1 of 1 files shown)


turtlemoji:
@huggingbot We should hold this since the layout and style changed in:
https://github.com/yearn/yearn-finance-v3/pull/694

PR #730, opened 2022-07-10
feat: `AmountInput` UI, disabled and error states

karelianpie:
## Description

Amount input component:

- Match colors and spacings to designs
- Disabled state
- Error state (red border)

## Motivation and Context

Match designs

## How Has This Been Tested?

Ran locally and observed the changes, refer to screenshots below

yearn_finance

yearn_finance

yearn_finance

### DIFF
--- a/src/client/components/app/AmountInput/index.tsx
+++ b/src/client/components/app/AmountInput/index.tsx
@@ -42,7 +42,12 @@ const StyledInput = styled.input<{ readOnly?: boolean; error?: boolean }>`
     }
   `}
 
-  ${({ error, theme }) => error && `color: ${theme.colors.txModalColors.error};`}
+  ${({ error, theme }) =>
+    error &&
+    `
+    color: ${theme.colors.txModalColors.error};
+    border: 1px solid red;
+  `}
 
   ${() => `
     ::-webkit-outer-spin-button,
--- a/src/client/components/app/AmountInput/index.tsx
+++ b/src/client/components/app/AmountInput/index.tsx
@@ -32,13 +32,13 @@ const StyledInput = styled.input<{ readOnly?: boolean; error?: boolean }>`
   ${({ readOnly, theme }) =>
     readOnly &&
     `
-    border: 1px solid ${theme.colors.textsVariant};
-    color: ${theme.colors.textsVariant};
+    border: 1px solid ${theme.colors.txModalColors.onBackgroundVariantColor};
+    color: ${theme.colors.txModalColors.onBackgroundVariantColor};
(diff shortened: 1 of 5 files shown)


src/client/components/app/AmountInput/index.tsx
```diff
@@ -32,17 +32,22 @@ const StyledInput = styled.input<{ readOnly?: boolean; error?: boolean }>`
...
-    border: 1px solid ${theme.colors.textsVariant};
-    color: ${theme.colors.textsVariant};
+    border: 1px solid ${theme.colors.txModalColors.onBackgroundVariantColor};
+    color: ${theme.colors.txModalColors.onBackgroundVariantColor};
     cursor: default;
     background: transparent;
 
     &::placeholder {
-      color: ${theme.colors.textsVariant};
+      color: ${theme.colors.txModalColors.onBackgroundVariantColor};
```
xgambitox, line 41: looking at the light color theme, seem they were using the correct color... I suppose you were testing with dark theme? Ideally both should use the same property, but if they are not consistent, we probably need to add a new one so that they match...

xgambitox (changes requested):
There might be inconsistent theme properties between light and dark that we need to solve...

src/client/components/app/AmountInput/index.tsx
karelianpie: Good catch, created an input placeholder prop in the theme and used it there

karelianpie: https://github.com/yearn/yearn-finance-v3/pull/730/commits/892c876e61e2116537469a164a77223d34a854c1 & https://github.com/yearn/yearn-finance-v3/pull/730/commits/80bf8a01fc985a6457fb0036d84f3a6f1638b22f

karelianpie:
@xgambitox did the style changes we've discussed yesterday (https://github.com/yearn/yearn-finance-v3/pull/730/commits/a288711cb3767b951ea9b518eb66a9cf5588ecd5) also noticed that in Figma when disabled the max button is not displayed (https://github.com/yearn/yearn-finance-v3/pull/730/commits/edd26d6d9b53cc93399683112ca17d21ce477910) see below.  Finally also added `cursor: not-allowed;` to make it clear it's disabled (https://github.com/yearn/yearn-finance-v3/pull/730/commits/8deff5478e955971e5d157387d7d0d1892017f9c):

Screenshot_13_7_22__10_09_am

Enabled:
Screen Shot 2022-07-13 at 10 27 32 am

xgambitox (approved):

PR #741, opened 2022-07-30
feat: Implement `SimpleTransaction` component

karelianpie:
## Description

Implement the Stake veYFI modal

## Motivation and Context

Current modal is using the old yearn style

yearn_finance

### DIFF
--- a/src/client/components/app/Transactions/StakeTx.tsx
+++ b/src/client/components/app/Transactions/StakeTx.tsx
@@ -1,9 +1,10 @@
 import { FC } from 'react';
 import { t } from 'i18next';
 
-import { AmountInput } from '@components/app';
 import { Box, Button, Dropdown } from '@components/common';
 import { formatAmount } from '@utils';
+
+import { AmountInput } from '../AmountInput';
 
 import { TxContainer } from './components/TxContainer';
 
--- a/src/client/components/app/Transactions/StakeTx.tsx
+++ b/src/client/components/app/Transactions/StakeTx.tsx
@@ -1,8 +1,8 @@
 import { FC } from 'react';
-import { t } from 'i18next';
 
 import { Box, Button, Dropdown } from '@components/common';
 import { formatAmount } from '@utils';
+import { useAppTranslation } from '@src/client/hooks';
 
 import { AmountInput } from '../AmountInput';
 
@@ -13,6 +13,7 @@ interface StakeTxProps {
 }
 
 export const StakeTx: FC<StakeTxProps> = (props) => {
+  const { t } = useAppTranslation('common');
   const { onClose } = props;
 
   const DUMMY_TOKEN = {
--- a/src/client/components/app/Transactions/StakeTx.tsx
+++ b/src/client/components/app/Transactions/StakeTx.tsx
@@ -1,0 +1,61 @@
+import { FC } from 'react';
+import { t } from 'i18next';
+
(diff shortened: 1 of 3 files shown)


src/client/components/app/Transactions/StakeTx.tsx
```diff
@@ -0,0 +1,61 @@
+import { FC } from 'react';
+import { t } from 'i18next';
```
xgambitox, line 2: import the `useAppTranslation` hook instead, and use it as
```const { t } = useAppTranslation('common');```

File: src/client/components/app/Transactions/StakeTx.tsx [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+
+import { AmountInput } from '@components/app';
```
xgambitox, line 4: I tend to be cautious when importing from the same folder that will export the component since circular dependencies are hard to debug. Use relative paths in that cases to avoid later on:   
```suggestion
import { AmountInput } from '../AmountInput';
```
note: you will also need to move it down our import groups since we have a lint rule

xgambitox:
Now that the UI is component is there, we should look into replacing the usage of the Transaction component in [Gauges/StakeTx](https://github.com/yearn/yearn-finance-v3/blob/feat/veYFI/src/client/components/app/Transactions/Gauges/StakeTx.tsx#L119) and [Gauges/Unstake](https://github.com/yearn/yearn-finance-v3/blob/feat/veYFI/src/client/components/app/Transactions/Gauges/UnstakeTx.tsx#L79) for this one (We should better rename this one to a more generic, like SimpleTransaction, so that it can be reused across other places). Some of the props used in the transaction component are no longer needed since this SimpleTransaction component doesnt render nothing about the _targetToken_.

Since we dont have specs for error or success state for the transaction, lets reuse the one used in Transaction component.

src/client/components/app/Transactions/StakeTx.tsx
karelianpie: Good catch, refactored with https://github.com/yearn/yearn-finance-v3/pull/741/commits/a6c4a934a78a59a0f92e863bbedde545413432df

src/client/components/app/Transactions/StakeTx.tsx
karelianpie: Refactored with https://github.com/yearn/yearn-finance-v3/pull/741/commits/95423b1a57b0eb31487b4169052c6c7b0e145478

karelianpie:
> Now that the UI is component is there, we should look into replacing the usage of the Transaction component in [Gauges/StakeTx](https://github.com/yearn/yearn-finance-v3/blob/feat/veYFI/src/client/components/app/Transactions/Gauges/StakeTx.tsx#L119) and [Gauges/Unstake](https://github.com/yearn/yearn-finance-v3/blob/feat/veYFI/src/client/components/app/Transactions/Gauges/UnstakeTx.tsx#L79) for this one (We should better rename this one to a more generic, like SimpleTransaction, so that it can be reused across other places). Some of the props used in the transaction component are no longer needed since this SimpleTransaction component doesnt render nothing about the _targetToken_.
> 
> Since we dont have specs for error or success state for the transaction, lets reuse the one used in Transaction component.

Would you be happy doing this in a different PR?

xgambitox (approved):

PR #744, opened 2022-08-17
feat: support portals zap service

xgambitox:
## Description

Replace Zapper zap service in favor of [portals.fi](https://app.portals.fi/) zap service

## Related Issue

https://github.com/yearn/yearn-sdk/pull/301

## Motivation and Context

Zapper discontinuing support for zaps.

## How Has This Been Tested?

Running app locally


src/client/components/app/Transactions/DepositTx.tsx
```diff
@@ -87,9 +87,7 @@ export const DepositTx: FC<DepositTxProps> = ({
       dispatch(
         TokensActions.setSelectedTokenAddress({
           tokenAddress:
-            !zapperEnabled && selectedVault.zapInWith === 'zapperZapIn'
-              ? selectedVault.token.address
-              : selectedVault.defaultDisplayToken,
+            !zapperEnabled && selectedVault.zapInWith ? selectedVault.token.address : selectedVault.defaultDisplayToken,
```
karelianpie, line 90: We could do here the same refactor as the sdk and rename `zapper`->`zap`, thoughts?

karelianpie (approved):
LGTM, it's worth it to remove the `zapper` keyword and replace it with `zap` for now IMO

src/client/components/app/Transactions/DepositTx.tsx
xgambitox: yeah agree, seems like best thing to do for now

